PR #2, opened 2019-09-30
Renames and changes for Python styling

mikemccourt:
I think I changed just enough to get the function directory mostly into shape.  I left some comments throughout, primarily in the `integrate_base.py` file, to explain some of what I did, but also what I might do going forward.

I think that what's there now is better, from a Pythonic perspective.  There are other things that I would change but I didn't want to push too much all at once.

Main focus of how I write code is to write "self-documenting" code -- code which should read exactly as it is to be executed and which, if changed, would yield appropriate changes to update the reading.  That means, different variable names under different circumstances, and minimal use of "math" letters (real names for things).

None of what I've done is required -- I just wanted to start the discussion.  Please leave comments in this PR and we can discuss how we want to design this repository going forward.  I am not the boss of this, just trying to give the same advice that I would give here at my company.  All of you own this so please make your opinions known.

Also, normally, I won't bother adding a bunch of people as reviewers -- usually just the people who were most recently editing the given files, or people who want to own a given topic that was being edited.

Eventually, I'll try to work on getting continuous integration up, and if we actually get the code to PEP8 compliance (or near, subject to our desires), I'll also investigate getting lint in place.


qmcpy/test/test_integrate.py
```diff
@@ -1,44 +1,48 @@
...
+    def test_keister_integrand(self):
+        abs_tol = .3
         dim = 2
-        fun = KeisterFun()
-        measureObj = measure().IIDZMeanGaussian(dimension=[dim],variance=[1 / 2])
-        distribObj = IIDDistribution(trueD=measure().stdGaussian(dimension=[dim]))
-        stopObj = CLTStopping(distribObj,nInit=16, absTol=absTol, alpha=.01, inflate=1.2)
-        sol, out = integrate(fun, measureObj, distribObj, stopObj)
+        integrand = KeisterFun()
+        measure = Measure().iid_zmean_gaussian(dimension=[dim], variance=[.5])
```
mikemccourt: Eventually, I think it would be better to have a `classmethod` in `Measure` which can create these objects ... I find the ```measure = Measure().iid_zmean_gaussian(dimension=[dim], variance=[.5])``` line kinda gross.

qmcpy/test/test_AsianCallFun.py
```diff
@@ -1,16 +1,24 @@
+import numpy as np
```
mikemccourt: I think this is much more common than importing the functions that you want.  Some people also do `from numpy import *`, but I am not personally a fan of that as that imports a bunch of stuff and can make it unclear when something is being defined in a local context versus when it is imported.

But I will leave this to you as I think it is viable either way.

qmcpy/algorithms/function/integrand_base.py
```diff
@@ -0,0 +1,162 @@
...
+# It requires a sense of the structure of the code base that shouldn't be necessary during development
+# But maybe I just feel that way because I like defining PYTHONPATH each time I run things
+# Feel free to do this as you want
+from .. import univ_repr
+
+# Eventually, could consider congealing defaults into a constants file
+DEFAULT_NOMINAL_VALUE = 0
+
+
+class IntegrandBase(object):
```
mikemccourt: Could also easily just be called `Integrand` ... I don't know why I threw the `Base` in there.

qmcpy/algorithms/function/__init__.py
```diff
@@ -1,69 +0,0 @@
-''' Originally developed in MATLAB by Fred Hickernell. Translated to python by Sou-Cheng T. Choi and Aleksei Sorokin '''
```
mikemccourt: I have definitely seen projects that use the `__init__.py` to do stuff, but usually it's just to surface content from the rest of the directory.  I would personally frown on the idea of important classes that will be used elsewhere being uniquely defined here.

mikemccourt: Also, putting historical comments in the code is odd -- generally the code is just considered alive and the GitHub history records how it got to where it is.

qmcpy/algorithms/distribution/__init__.py
```diff
@@ -18,15 +18,15 @@ def __init__(self, accepted_measures, trueD=None, distrib_data=None):
         super().__init__()
         # Abstract Properties
         self.distrib_data = distrib_data
-        self.trueD = trueD
```
mikemccourt: I use PyCharm (I think I saw one of you using VSCode) and when I save a file in PyCharm it automatically applies my file standards (which in this case is remove trailing whitespace).  It's also standard for PEP8.  When I get lint in place I'll make a lot of edits like this.

qmcpy/algorithms/accum_data/MeanVarData.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, nf: int) -> None:
         self.nSigma = zeros(nf) # number of samples used to compute the sample standard deviation
         self.nMu = zeros(nf)  # number of samples used to compute the sample mean
 
-    def update_data(self, distrib_obj: DiscreteDistribution, fun_obj: Fun) -> None:
+    def update_data(self, distrib_obj: DiscreteDistribution, fun_obj: IntegrandBase) -> None:
```
mikemccourt: I'm not someone who likes typing in Python, but that's a personal preference.  Facebook is all about it now, Google seems ambivalent, Uber doesn't do it ... opinions across the spectrum.  I would say, though, that if you want to have these type suggestions that it should be done consistently throughout.  Nothing wrong with starting one place and continuing, but it should be decided whether this is desired or not.

.gitignore
```diff
@@ -18,4 +18,7 @@ qmcpy/workouts/Outputs/*.png
 qmcpy/__pycache__/
 qmcpy/sphinx/_build/
 qmcpy/sphinx/_static/
-qmcpy/sphinx/_templates/
\ No newline at end of file
+qmcpy/sphinx/_templates/
+
+# Virtual environment stuff
+venv
```
mikemccourt: Not sure if people are using virtual environments, but I generally use it when I have a new project.  We could consider suggesting it to customers in the `README`.

qmcpy/algorithms/distribution/__init__.py
```diff
@@ -51,13 +51,13 @@ def __setitem__(self,i,val): self.distrib_list[i]=val
     def __repr__(self): return univ_repr(self,'distrib_list')
 
 
-class Measure():
+class Measure(object):
```
mikemccourt: The standard is to always inherit from something (the base being `object`).  It's not technically required, but not doing so is frowned upon.

qmcpy/algorithms/function/asian_call_integrand.py
```diff
@@ -0,0 +1,50 @@
+''' Originally developed in MATLAB by Fred Hickernell. Translated to python by Sou-Cheng T. Choi and Aleksei Sorokin '''
```
mikemccourt: Putting historical comments in the code is odd -- generally the code is just considered alive and the GitHub history records how it got to where it is.  If you want to make a permanent historical statement, putting it in a separate text file elsewhere would be preferable (as this file could be replaced by something else tomorrow).

qmcpy/algorithms/function/asian_call_integrand.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class AsianCallFun(IntegrandBase):
+    """
+    Specify and generate payoff values of an Asian Call option
+
+    TODO - Come up with better, more interpretable, lowercase names for these parameters
+
+    """
+    def __init__(self, bm_measure=None, volatility=.5, S0=30, K=25, nominal_value=None):
```
mikemccourt: I mentioned this somewhere else, but variable names should be intelligible and lowercase.  This is math code, so some extra flex is permissible (such as just using the letter `K` rather than saying `kernel_matrix`), but not as arguments in a function (only within the function as a local variable).

qmcpy/algorithms/function/asian_call_integrand.py
```diff
@@ -0,0 +1,50 @@
...
+        super().__init__(nominal_value)
+        self.bm_measure = bm_measure
+        self.volatility = volatility
+        self.S0 = S0
+        self.K = K
+        self.dimFac = 0
+
+        if self.bm_measure:
+            num_bm = len(bm_measure)
+            self.fun_list = [AsianCallFun() for i in range(num_bm)]
```
mikemccourt: This just looks so weird -- recursion in Python is generally frowned upon (though permitted) but creating a class populated by itself is super odd.  My suggestion, eventually, would be to just create a subclass called `AsianCallFunWithoutBM`, or maybe something better than that.

qmcpy/algorithms/function/asian_call_integrand.py
mikemccourt: Also, docstrings should be `"""` not `'''`

qmcpy/algorithms/function/asian_call_integrand.py
alegresor: I agree that it looks strange. 
It's a little bit of a hacky way to allow for mulit-dimensional problems and to hide their construction from the user. 
If there is an easier way to implement something like this that may be helpful to us and users. 
Note: there are similar constructions for Measure and DiscreteDistributuion.

qmcpy/algorithms/function/asian_call_integrand.py
alegresor: These have been changed in the most recent commit. 
s0 -> initial_price
k -> strike_price

qmcpy/algorithms/function/asian_call_integrand.py
alegresor: The line ''' Originally developed in MATLAB ...''' has been removed from all files in the most recent commit. 
As we go through and make files PEP8 compatible we'll change ''' to """
Files in algorithms/accum_data and algorithms/stop have been updated to have better docstrings, variable names (lower case), and google styling. 
Throughout the week we'll get the rest of the files looking nice

qmcpy/algorithms/distribution/__init__.py
alegresor: Measure is now an abstract class (inherits from ABC) with similar construction to DiscreteDistribution

.gitignore
alegresor: I use pip with virualenv and virtualenvwrapper. I think everyone is set up with a virtual environment for the project, but we can confirm at the next meeting. 

I added the following to the end of qmcpy/README.md:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
# Tips
It may be helpful to create a virtual environment for this project<br>
Packages and versions are compiled in requirements.txt<br>
In order for the software to find packages, make sure .../qmcpy/ is in your path

qmcpy/algorithms/accum_data/MeanVarData.py
alegresor: we've been adding typing to files as they get updated documentation.
I'm not sure if someone who wants to contribute to our software would be familiar and ready to use typing.
Maybe it is a good idea in the abstract classes but not in the concrete implementations? Would this break consistency? Just typing our softwares objects?

qmcpy/algorithms/function/__init__.py
alegresor: Should we move the abstract classes to  files of the same name?
For example, should we put the Integrand abstarct class at qmcpy/algorithms/integrand/Integrand.py? maybe .../IntegrandBase.py?

qmcpy/algorithms/function/integrand_base.py
alegresor: Renamed to Integrand in the most recent commit

qmcpy/test/test_integrate.py
alegresor: updated in most recent commit
Measure in now an abstract class that has instances in algorithms/distributuions/Measures.py

New lines reads:
    from algorithms.distributions.Measures import IIDZeroMeanGaussian
    measure = IIDZeroMeanGaussian(...)

qmcpy/algorithms/distribution/__init__.py
mikemccourt-sigopt: Is there some functionality in `ABC` that you are using here.  I didn't immediately see anything that actually required it.  I would, in general, avoid using `ABC` unless there's a specific functionality that's required (which can happen but is generally not the case in Python).

qmcpy/algorithms/accum_data/MeanVarData.py
mikemccourt: That's part of the reason I don't generally bother doing it at all -- it's a lot to keep up throughout the code base for, what is often, almost no real benefit.

Given that, one advantage of limiting merge control is that before you let external contributors push content to master they have to satisfy your expectations.  If an expectation is that types are defined, then you can enforce that and hold up a PR until it's fixed.

qmcpy/algorithms/distribution/__init__.py
mikemccourt: Is there some functionality in ABC that you are using here. I didn't immediately see anything that actually required it. I would, in general, avoid using ABC unless there's a specific functionality that's required (which can happen but is generally not the case in Python).
@mikemccourt

mikemccourt: Also, sorry, I accidentally posted something on here on my SigOpt account -- I'll need to be more careful about that.

alegresor: It was changed to an abstract (abc) class to allow for behind-the-scenes multi-dimensional measure construction (similar to AsianCallOption and the Abstract class DiscreteDistribution)
Just makes it easier for contributes to add their own measures without having to worry about creating measure objects that act as lists of measure objects

mikemccourt: I don't think that's actually true -- I think if you just inherit from `object` and `raise NotImplementedError` in undefined functions it will do the same thing (without the more complicated abstract structure in place).

In my mind, the only 2 cases where `ABC` is actually useful are

* in circumstances where instantiation of a class is extremely expensive and you want to throw an error at instantiation time if someone tries to instantiate a pure virtual class (as opposed to simply erroring when it gets used), or
* if you are leveraging duck typing throughout the code and want to replace `hasattr` calls with `isinstance` or `issubclass` calls.

Beyond that, I don't think there's too much benefit to using `ABC`.  Again, I'm not going to say what you guys need to do, but I do think that the use of `ABC` is pretty uncommon unless a programmer coming from a Java background.

mikemccourt:
I'm not sure I'm going to get to this merge before my plane leaves. Sorry
for the delay. I'll try to get to it soon. Please don't let this delay
anything you guys are working on.

mikemccourt:
As I'm going through these merge conflicts I realized you guys already made all these changes.  Very nice.  Thank you for your incredibly efficient work.  I'll close this now.

PR #10, opened 2019-10-13
Refactor distributions

alegresor:
Eliminates the need for IID Distribution and QuasiRandom distribution types. 
A user would now choose a discrete_distribution and a true_distribution.
The variable transformation  is no longer preformed in the integrand but is now preformed in the true_distribution.
A short example is below, but lots more examples can be found in the workouts

`
from qmcpy.discrete_distribution import IIDStdUniform
from qmcpy.true_distribution import Gaussian

discrete_distrib = IIDStdUniform()
true_distrib = Gaussian(dimension=[3], mean=1, variance=1/2)
true_distrib.transform_generator(discrete_distrib)
'''
The above class method is called by the integrate method in place of transform_function.
This class method will give the true distributuion for each integrand a generator for samples that mimic the true_distribution. 
e.g. we can now call true_distrib[0].gen_distribution(5,4) to get a 5x4x3 matrix of samples for the 1st integrand
`

Both fast and long unittests are passing

There is not much documentation for qmcpy/discrete_distribution/ or qmcpy/true_distribution/

Not sure if this is the best approach, but I think it's more clear than what we had with distributions and measures.


fjhickernell:
Thanks for thinking this through, but sometimes we integrate with respect to a measure, not a distribution.  So maybe we should have true_measure class, not a true_distribution class.  What is the output of true_distrib.transform?

alegresor:
"Thanks for thinking this through, but sometimes we integrate with respect to a measure, not a distribution. So maybe we should have true_measure class, not a true_distribution class. What is the output of true_distrib.transform?"
--
I changed the name of the True Distribution folder/class to True Measure in the latest commit. 

true_measure.transform_generator(discrete_distrib) creates the method gen_distribution for each true_measure. It doen't output or return anything, just creates the gen_distribution method which will generate samples mimicing the true_measure.

lmatar:
https://doodle.com/poll/wixexwqfun4gu4sn <https://doodle.com/poll/wixexwqfun4gu4sn>

This is the link for the doodle to find a time that works for all of us.  I made the settings for the next two weeks so there is a greater chance to find a good date that works for all of us.

Issue #197, opened 2020-10-31
Sobol' generator random seeding prints warning


PieterjanRobbe:
I am getting a warning message printed on the screen when the Sobol' generator is accidentally called with a random seed "0". Here's an example:

```
>>> sob = Sobol(100, seed=339)
>>> sob.gen_samples()
seed must be in [0,9223372036854754679] but 0. Using seed=7
[...]
```

The error is thrown here: 

https://github.com/QMCSoftware/QMCSoftware/blob/24e506a65d48f29bae1f028b7ef369a924d4a6d5/qmcpy/discrete_distribution/c_lib/MRG63k3a.c#L96

And is caused by this line:

https://github.com/QMCSoftware/QMCSoftware/blob/24e506a65d48f29bae1f028b7ef369a924d4a6d5/qmcpy/discrete_distribution/sobol/sobol.py#L196

I think this is because `random.randint(a, b)` generates random integers with `a` and `b` included.

```
>>> random.seed(339)
>>> seeds = random.randint(0, 100000, size=100, dtype=uint64)
>>> 0 in seeds
True
```

PR #198, opened 2020-11-01
Fix for #197

PieterjanRobbe:


PieterjanRobbe:
I noticed that the Travis CI test fails on Windows, is this something you expected?

alegresor:
We have seen this type of error before, I'll put in a patch for the failing
doctest soon.

PR #223, opened 2021-01-30
not fixing a seed globally

jungtaekkim:
Hi,

It works in my environment.

If it is okay, I will push some more commits that will fix other samplers as well.

Best regards,
Jungtaek.


alegresor:
Thanks for pointing this out!
I pushed a change to the develop branch, let me know if the issue persists.

PR #260, opened 2021-07-09
LR.py

CDHJ2000:
Claude Hall's program to integrate logistic regression with data


alegresor:
I've implemented this in the QMCPy `Integrand` named  `BayesianLRCoeffs`, see https://github.com/QMCSoftware/QMCSoftware/blob/develop/qmcpy/integrand/bayesian_lr_coeffs.py

PR #293, opened 2022-07-01
Doc req updates

alegresor:
- consolidate dependencies to a single `environment.yml` file compatible across all platforms with Python 3.9
- update GitHub Actions and ReadTheDocs to compile based on new `environment.yml` file
- note that we remove LatNet Builder from the developer environment in order to use Python 3.9. When LatNet Builder updates to support Python >= 3.9 we will add it back


sou-cheng-choi:
Dear Aleksei, I have tested in on Mac M1. After setting up a new virtual environment with the much cleaner environment.yml, all tests are passing. For the make targets such as "doc_html", I got the following erorr message:

`
# Make Directries
# READMEs --> RST
make: pandoc: No such file or directory
make: *** [_doc] Error 1
`

So I proceeded to install pandoc by "conda install -c conda-forge pandoc". Afterwards, all documentation generation tasks became successful.

PR #308, opened 2023-01-05
Updated plot method

Baronlegend27:
Updated the plot method

### DIFF
--- a/qmcpy/discrete_distribution/_discrete_distribution.py
+++ b/qmcpy/discrete_distribution/_discrete_distribution.py
@@ -52,8 +52,38 @@ class DiscreteDistribution(object):
         """ ABSTRACT METHOD to evaluate pdf of distribution the samples mimic at locations of x. """
         raise MethodImplementationError(self, 'pdf')
 
-    def plot(self, n, d_vertical, d_horizontial):
-        print("Hello, world")
+    def plot(self, n, d_vertical=0, d_horizontial =1):
+        """
+        ABSTRACT METHOD for ploting a distrubution, you will also need to import matplotlib.pyplot as plt
+
+        Args:
+            n (int): n is the number of samples that will be calculated 
+
+            d_vertical (int): d_vertical is the index of points that will be plotted on the vertical axis. See below
+            
+
+            d_horizontial (int): d_horizontial is the index of points that will be plotted as the horizontial axis. See below
+
+        d_horizontial and d_vertical explanation:
+            Both of theses numbers act as an index in an array of dimensions for the given object. 
+            The chosen number for d_horizontial or d_vertical will index a given dimension. With the d_horizontial the dimension will be on the 
+            x-axis and the d_vertical index will then become the y-axis. 
+
+            
+
+        The plot method should be defined similar to :
+            fig, ax = plt.subplots(nrows=1, ncols=1, figsize=(16, 6))
+            samples = self.gen_samples(n)
+            ax.scatter(samples[:, d_vertical], samples[:, d_horizontial])
+            plt.show()
+
+        
+        This method does not return anything it just plots a scatter plot.
+
+
+        """
+        
+        raise MethodImplementationError(self, 'plot')
 
     def spawn(self, s=1, dimensions=None):
         """
--- a/qmcpy/discrete_distribution/_discrete_distribution.py
+++ b/qmcpy/discrete_distribution/_discrete_distribution.py
@@ -1,5 +1,6 @@
 from ..util import ParameterError, MethodImplementationError, _univ_repr, DimensionError
 from numpy import *
+import matplotlib.pyplot as plt
 
 
 class DiscreteDistribution(object):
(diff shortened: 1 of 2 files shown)


Baronlegend27 (commented):
I made the abstract method called plot. I am pretty sure everything is correct, however I wrote a lot of documentation for the method.

qmcpy/discrete_distribution/_discrete_distribution.py
```diff
@@ -52,6 +52,39 @@ def pdf(self, x):
         """ ABSTRACT METHOD to evaluate pdf of distribution the samples mimic at locations of x. """
         raise MethodImplementationError(self, 'pdf')
 
+    def plot(self, n, d_vertical=0, d_horizontial =1):
+        """
+        ABSTRACT METHOD for ploting a distrubution, you will also need to import matplotlib.pyplot as plt
```
alegresor, line 57: Although this method is in the abstract class, this is not an abstract method. 

You should `impot matplotlib.pyplot as plt` in the routine. If you put this import in a try-except then you can throw a friendly error for the user who tries to call this routine without having matplotlib installed.

qmcpy/discrete_distribution/_discrete_distribution.py
```diff
@@ -52,6 +52,39 @@ def pdf(self, x):
...
+
+        Args:
+            n (int): n is the number of samples that will be calculated
```
alegresor, line 60: maybe "plotted" instead of "calculated"

qmcpy/discrete_distribution/_discrete_distribution.py
```diff
@@ -52,6 +52,39 @@ def pdf(self, x):
         """ ABSTRACT METHOD to evaluate pdf of distribution the samples mimic at locations of x. """
         raise MethodImplementationError(self, 'pdf')
 
+    def plot(self, n, d_vertical=0, d_horizontial =1):
+        """
+        ABSTRACT METHOD for ploting a distrubution, you will also need to import matplotlib.pyplot as plt
+
+        Args:
+            n (int): n is the number of samples that will be calculated 
+
```
alegresor, line 61: we don't leave white space between arguments

qmcpy/discrete_distribution/_discrete_distribution.py
```diff
@@ -52,6 +52,39 @@ def pdf(self, x):
...
+            d_vertical (int): d_vertical is the index of points that will be plotted on the vertical axis. See below
```
alegresor, line 62: you don't need the "see below"

qmcpy/discrete_distribution/_discrete_distribution.py
```diff
@@ -52,6 +52,39 @@ def pdf(self, x):
...
+            
+
+            d_horizontial (int): d_horizontial is the index of points that will be plotted as the horizontial axis. See below
+
+        d_horizontial and d_vertical explanation:
+            Both of theses numbers act as an index in an array of dimensions for the given object. 
+            The chosen number for d_horizontial or d_vertical will index a given dimension. With the d_horizontial the dimension will be on the 
+            x-axis and the d_vertical index will then become the y-axis.
```
alegresor, line 70: you probably don't need this. I think the the strings for `d_vertical` and `d_horizontal` are sufficient. please fix the spelling of horizontal.

qmcpy/discrete_distribution/_discrete_distribution.py
```diff
@@ -52,6 +52,39 @@ def pdf(self, x):
...
+            The chosen number for d_horizontial or d_vertical will index a given dimension. With the d_horizontial the dimension will be on the 
+            x-axis and the d_vertical index will then become the y-axis. 
+
+            
+
+        The plot method should be defined similar to :
+            fig, ax = plt.subplots(nrows=1, ncols=1, figsize=(16, 6))
+            samples = self.gen_samples(n)
+            ax.scatter(samples[:, d_vertical], samples[:, d_horizontial])
+            plt.show()
```
alegresor, line 78: Move this into the method so it executes when called

qmcpy/discrete_distribution/_discrete_distribution.py
```diff
@@ -52,6 +52,39 @@ def pdf(self, x):
...
+
+        
+        This method does not return anything it just plots a scatter plot.
```
alegresor, line 81: return the `fig,ax` in case the user wants to further modify our baseline plots.

qmcpy/discrete_distribution/_discrete_distribution.py
```diff
@@ -52,6 +52,39 @@ def pdf(self, x):
...
+
+
+        """
+        
+        raise MethodImplementationError(self, 'plot')
```
alegresor, line 86: the code for plotting that you currently have in the docstring goes here

alegresor (changes requested):
@Baronlegend27 This is a good start, the documentation for the method looks nice, I left some suggestions in this review. You still need to implement the method. This is not an abstract method but rather a method in the abstract class. Any subclass will inherit this `plot` method so you do not need to rewrite this method for every discrete distribution.

qmcpy/discrete_distribution/_discrete_distribution.py
```diff
@@ -1,5 +1,6 @@
 from ..util import ParameterError, MethodImplementationError, _univ_repr, DimensionError
 from numpy import *
+import matplotlib.pyplot as plt
```
alegresor, line 3: Move this into a try except within the method

qmcpy/discrete_distribution/_discrete_distribution.py
```diff
@@ -57,33 +58,22 @@ def plot(self, n, d_vertical=0, d_horizontial =1):
...
-        d_horizontial and d_vertical explanation:
-            Both of theses numbers act as an index in an array of dimensions for the given object. 
-            The chosen number for d_horizontial or d_vertical will index a given dimension. With the d_horizontial the dimension will be on the 
-            x-axis and the d_vertical index will then become the y-axis. 
-
-            
-
-        The plot method should be defined similar to :
+        """
+        try:
```
alegresor, line 68: the try except should only try to `import matplotlib.pyplot as plt` and except an error that it is not installed.

qmcpy/discrete_distribution/_discrete_distribution.py
```diff
@@ -57,33 +58,22 @@ def plot(self, n, d_vertical=0, d_horizontial =1):
...
             fig, ax = plt.subplots(nrows=1, ncols=1, figsize=(16, 6))
```
alegresor, line 69: figsize=(16,6) will be quite wide for a 1 column by 1 row plot. Maybe try something like (5,5) or play around with these values

qmcpy/discrete_distribution/_discrete_distribution.py
```diff
@@ -57,33 +58,22 @@ def plot(self, n, d_vertical=0, d_horizontial =1):
...
             samples = self.gen_samples(n)
             ax.scatter(samples[:, d_vertical], samples[:, d_horizontial])
```
alegresor, line 71: switch d_vertical and d_horizontal. The first argument is the x axis locations (horizontal) and the second argument is the y axis locations (vertical)

qmcpy/discrete_distribution/_discrete_distribution.py
```diff
@@ -57,33 +58,22 @@ def plot(self, n, d_vertical=0, d_horizontial =1):
...
             plt.show()
```
alegresor, line 72: don't show the plot, just return the fig,ax (as you have)

alegresor (changes requested):
Looks better! Left a few more suggestions on how to improve the method. 

You may also allow a user to pass in an `ax` argument which defaults to `None`. If an `ax` is passed in, you may just use that one to make the scatter plot on. If `ax is None` then you can create the `fig,ax` for the user. 

Add some commands to pretty up the plot e.g. ones that set the aspect ratio, axis labels, axis ticks, axis limits

alegresor (approved):

PR #324, opened 2023-07-15
Matern kernel extension of Gaussian true measure

thegman108:
matern.py calculates the covariance matrix using a Matern kernel as described in https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.kernels.Matern.html.


qmcpy/true_measure/matern.py
```diff
@@ -0,0 +1,89 @@
+from .gaussian import Gaussian
+from ..discrete_distribution._discrete_distribution import DiscreteDistribution
+from ._true_measure import TrueMeasure
+from ..discrete_distribution.lattice.lattice import Lattice
+from ..util import DimensionError, ParameterError
+import numpy as np
```
alegresor, line 6: We usually just do `from numpy import *`

qmcpy/true_measure/matern.py
```diff
@@ -0,0 +1,89 @@
...
+from scipy.stats import norm, multivariate_normal
+from sklearn import gaussian_process as gp
+
+class Matern(Gaussian):
+    """
+    A normal measure using a Matern kernel as the covariance matrix.
+    >>> import numpy as np
```
alegresor, line 13: No need to re-import here

qmcpy/true_measure/matern.py
```diff
@@ -0,0 +1,89 @@
...
+            sampler (DiscreteDistribution/TrueMeasure): A 
+                discrete distribution from which to transform samples or a
+                true measure by which to compose a transform. 
+            x_values (array): The positions of points on a metric space.
+            length_scale (float): Determines "peakiness", or how correlated 
+                two points are based on their distance.
+            nu (float): The "smoothness" of the Matern function. e.g. nu=1.5
+                implies a once-differentiable function, while nu=2.5 implies twice
+                differentiability. Meanwhile, when nu=0.5, the Matern kernel equals
+                the RBF kernel, while when nu=inf, it equals the absolute exponential
```
alegresor, line 58: I thought Matern approached RBF as nu went to Inf. I think nu=1/2 implies less smoothness than nu=3/2

qmcpy/true_measure/matern.py
```diff
@@ -0,0 +1,89 @@
...
+                kernel. Note that nu values not in [0.5, 1.5, 2.5, inf] will be 
+                ~10x slower to run.
+            variance (float): The variance (or the diagonal elements of the
+                covariance matrix) at each point.
```
alegresor, line 62: Shouldn't these just be computed using the Matern kernel?

qmcpy/true_measure/matern.py
```diff
@@ -0,0 +1,89 @@
+from .gaussian import Gaussian
+from ..discrete_distribution._discrete_distribution import DiscreteDistribution
+from ._true_measure import TrueMeasure
+from ..discrete_distribution.lattice.lattice import Lattice
+from ..util import DimensionError, ParameterError
+import numpy as np
+from scipy.stats import norm, multivariate_normal
+from sklearn import gaussian_process as gp
```
alegresor, line 8: we don't want to introduce a new dependency in sklearn. Try to just implement it yourself and make sure they match

qmcpy/true_measure/matern.py
```diff
@@ -0,0 +1,89 @@
...
+        self.x_values = x_values
+        self.length_scale = length_scale
+        self.nu = nu
+        self.variance = variance
+        kernel = gp.kernels.Matern(length_scale = length_scale, nu=nu)
+        tnp = np.array([x_values]).T
+        #print(tnp)
+        self.covariance = variance * kernel.__call__(tnp) #takes array of size (# samples) x (# data points)
+        #print(covariances)
+        sampler_out = sampler.spawn(1, len(x_values))[0] #new version of sampler
```
alegresor, line 82: Don't spawn a new sampler, just use the one passed in

qmcpy/true_measure/matern.py
```diff
@@ -0,0 +1,89 @@
...
+        super().__init__(sampler_out, mean=mean, covariance=self.covariance, decomp_type=decomp_type)
+
+    def _spawn(self, sampler):
+        return Matern(sampler.spawn(1, len(self.x_values))[0], self.x_values, length_scale=self.length_scale, nu=self.nu, variance=self.variance,
```
alegresor, line 86: the sampler passed into this function is already a spawn, no need to spawn another one

qmcpy/true_measure/matern.py
```diff
@@ -0,0 +1,89 @@
...
+        if not (isinstance(sampler, DiscreteDistribution) or isinstance(sampler, TrueMeasure)):
+            raise ParameterError("sampler input should either be a DiscreteDistribution or TrueMeasure.")
+        self.x_values = x_values
+        self.length_scale = length_scale
+        self.nu = nu
+        self.variance = variance
+        kernel = gp.kernels.Matern(length_scale = length_scale, nu=nu)
+        tnp = np.array([x_values]).T
+        #print(tnp)
+        self.covariance = variance * kernel.__call__(tnp) #takes array of size (# samples) x (# data points)
```
alegresor, line 80: I see, the variance is a factor to multiply the kernel matrix by along the diagonal. You should clarify this in the documentation.

alegresor, line 80: This should be straightforward to implement without calling sklearn, see https://en.wikipedia.org/wiki/Mat%C3%A9rn_covariance_function#Simplification_for_specific_values_of_%CE%BD

alegresor (changes requested):
Good idea, just implement the matern kernel without calling sklearn so we don't have to add another dependency. Some other minor comments are included as well.

qmcpy/true_measure/matern.py
thegman108: Right, I swapped those, thanks for noticing

qmcpy/true_measure/matern.py
```diff
@@ -0,0 +1,140 @@
+from .gaussian import Gaussian
+from ..discrete_distribution._discrete_distribution import DiscreteDistribution
+from ._true_measure import TrueMeasure
+from ..discrete_distribution.lattice.lattice import Lattice
+from ..util import DimensionError, ParameterError
+from numpy import *
+#from scipy.stats import norm, multivariate_normal
```
alegresor, line 7: delete if not using

qmcpy/true_measure/matern.py
```diff
@@ -0,0 +1,140 @@
...
+            sampler (DiscreteDistribution/TrueMeasure): A 
+                discrete distribution from which to transform samples or a
+                true measure by which to compose a transform. 
+            x_values (array): The positions of points on a metric space.
+            length_scale (float): Determines "peakiness", or how correlated 
+                two points are based on their distance.
+            nu (float): The "smoothness" of the Matern function. e.g. nu=1.5
+                implies a once-differentiable function, while nu=2.5 implies twice
+                differentiability. Meanwhile, when nu=0.5, the Matern kernel equals
+                the RBF kernel, while as nu approaches inf, it equals the absolute
```
alegresor, line 86: still flipped

qmcpy/true_measure/matern.py
```diff
@@ -0,0 +1,140 @@
...
+            nu (float): The "smoothness" of the Matern function. e.g. nu=1.5
+                implies a once-differentiable function, while nu=2.5 implies twice
+                differentiability. Meanwhile, when nu=0.5, the Matern kernel equals
+                the RBF kernel, while as nu approaches inf, it equals the absolute 
+                exponential kernel. Note that nu values not in [0.5, 1.5, 2.5, inf] 
+                will be ~10x slower to run.
+            variance (float): The variance (or the diagonal elements of the
+                covariance matrix) of the distribution at each point. Implemented
+                as a constant factor of the Matern covariance. The default 
+                value is 1.0.
```
alegresor, line 92: I would call this a scaling factor since it multiplies *all* values of the kernel matrix by a constant.

alegresor (changes requested):
You should accommodate multidimensional sampling locations and provide a doctest showing this capability.

thegman108:
> You should accommodate multidimensional sampling locations and provide a doctest showing this capability.

As in a multidimensional x_values array? Or something else?

alegresor:
Yes. x_values should accommodate an (d x s) ndarray. Currently it is
restricted to a length d vector

alegresor (changes requested):
Looks good. Don't allow points or distance. Just have the user pass in points that is a N x s array. s can be 1. Or if the user just passes in a length N vector then just reshape into a N x 1. This will make for less conditionals and a simplified API

thegman108:
> Looks good. Don't allow points or distance. Just have the user pass in points that is a N x s array. s can be 1. Or if the user just passes in a length N vector then just reshape into a N x 1. This will make for less conditionals and a simplified API

fixed (removed distance option)

alegresor (approved):
Nice job! Thanks for accommodating the suggestions

PR #330, opened 2023-08-14
Fast lattice

sou-cheng-choi:
Speed up parallel MPS Lattice generation by a factor of 3.6 compared to serial.


sou-cheng-choi (commented):
Passed all tests and verified speed up factor.

PR #331, opened 2023-08-14
Fast lattice

sou-cheng-choi:
Faster MPS Lattice Generation.


alegresor (changes requested):
Look good! It seems the automatic test asserting the parallel time is faster than the serial time is failing.

alegresor (approved):

PR #351, opened 2024-07-03
Develop qmc finance

ggraoigr:
Pull request to merge with develop branch


qmcpy/integrand/LookBackOption.py
```diff
@@ -0,0 +1,91 @@
...
+        self.volatility = volatility
+        self.stock_values =self.start_price*np.exp((self.interest_rate-0.5*self.volatility**2)*self.t+self.volatility*self.true_measure.gen_samples(self.n))
+        super(LookBackOption,self).__init__(dimension_indv=1,dimension_comb=1,parallel=False)    
+        
+    def f(self, x, periodization_transform='NONE', compute_flags=None, *args, **kwargs):
+        """Overrides the parent's method. Refer to the parent class method for details of original method."""
+        self.n = len(x)
+        self.stock_values = self.start_price * np.exp((self.interest_rate - 0.5 * self.volatility**2) *
+                                                    self.t + self.volatility * self.true_measure.gen_samples(self.n))
+        return super().f(x, periodization_transform, compute_flags, *args, **kwargs)
```
alegresor, line 47: Do not override this method. 
The transform to a geometric brownian motion is already set to `self.s` in the `g` method below

qmcpy/stopping_criterion/.ipynb_checkpoints/_cub_qmc_ld_g-checkpoint.py
alegresor, line 1: How is this different than the existing `_CubQMCLDG` class?
Please remove if it is a duplicate or merge changes into the existing class.

qmcpy/stopping_criterion/Fixed_multilevel.py
```diff
@@ -0,0 +1,106 @@
...
+        self.discrete_distrib = self.integrand.discrete_distrib
+        # Verify Compliant Construction
+        allowed_levels = ['fixed-multi']
+        allowed_distribs = [LD]
+        allow_vectorized_integrals = False
+        super(CubQMCFML,self).__init__(allowed_levels, allowed_distribs, allow_vectorized_integrals)
+        
+    def integrate(self):
+        # Construct AccumulateData Object to House Integration Data
+        self.data = MLQMCData(self, self.integrand, self.true_measure, self.discrete_distrib,
```
alegresor, line 60: I'm not sure how this stopping criterion works. The `MLQMCData` object is for adaptive multi-level problems and the adaptive mechanisms are used below e.g. in `self.data._add_level()`. However, this seems to be written for fixed multi-level problems? 

Do you have a writeup or reference on how this works?

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
```diff
@@ -5,6 +5,9 @@
 from numpy import *
 from time import time
 import warnings
+import matplotlib.pyplot as plt
```
alegresor, line 8: move this to within the method performing the plotting. This makes it so matplotlib is only imported if the user wants to plot something, but does not make it a requirement of QMCPy.

alegresor (changes requested):
Thanks for putting this together all! 

Left a number of comments below. 

I think the american option needs some extra thought as to how we should update previous samples in a stopping criterion. 

Please provide a reference for the fixed multi-level stopping criterion. I think the current algorithm is actually adaptive. 

The lookback option seems fine. 

The density estimation method should be moved to the accumulate data class, be refactored to make plotting optional, and be allowed to take in samples at which we evaluate the combined discrete/continuous KDE for the user. 

I'll take a closer look at these files, including the notebooks, when we are closer to merging. 

If you'd like to make a separate PR with just the density estimation and lookback options parts I think we can get that merged quicker than the american option and fixed multi-level parts

qmcpy/integrand/AmericanOption.py
```diff
@@ -0,0 +1,115 @@
...
+        values=payoff[:,self.observations-1] #Option values after comparing payoffs with linear regression
+        #print("values",values)
+        for j in range(self.observations-1,0,-1):
+            in_money_values=payoff[:,j]>0
+            x=self.stock_values[in_money_values,j-1]
+            y=values[in_money_values]
+            A = np.vstack([x, np.ones(len(x))]).T
+            m, c = np.linalg.lstsq(A, y, rcond=None)[0] #use multilinear regression
+            hold_lr=m*self.stock_values[in_money_values,j-1]+c #lr of holding value one more time step
+            exercise=in_money_values
```
rvare, line 64: I want to note that on line 64 which is `exercise=in_money_values` could have potential problems later on. This statement does not make a copy the list but instead has `exercise` point to the same list that `in_money_values` also points to. At the moment it seems that the code works as it should due to pure luck, but this could break at a later time. However, if this was intended, then disregard this comment.

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
```diff
@@ -156,3 +159,36 @@ def set_tolerance(self, abs_tol=None, rel_tol=None):
...
         if abs_tol != None: self.abs_tol = abs_tol
         if rel_tol != None: self.rel_tol = rel_tol
+
+    def density_estimation(self):
+        approx_solution, data = self.integrate()
+        g=data.datum[0].y_val
+        a=min(g)
+        b=max(g)
+        print("Lower bound of estimation interval (a) is:",a)
+        print("Upper bound of estimation interval (b) is:",b)
```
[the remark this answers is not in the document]
rvare: I will delete these statements.

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
```diff
@@ -156,3 +159,36 @@ def set_tolerance(self, abs_tol=None, rel_tol=None):
...
+        unique_values, counts = unique(g, return_counts=True)
+        repeated_values = unique_values[counts > 1]# Find elements that are repeated (count > 1)
+        values_cont=unique_values[counts==1]
+        counts=counts[counts > 1]/len(g) #probability of each discrete value
+        prob=len(values_cont)/len(g) #probability of being continous
+        kde=sc.stats.gaussian_kde(values_cont.T)
+        z=linspace(a,b,100)
+            
+        pdf_est=kde.evaluate(z.T)*prob
+        fig, ax1 = plt.subplots()
```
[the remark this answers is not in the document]
rvare: What I'll do is make an if statement and change the method header to have a Boolean flag to make plotting optional.

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
```diff
@@ -156,3 +159,36 @@ def set_tolerance(self, abs_tol=None, rel_tol=None):
...
+        ax1.plot(z, pdf_est, color=color)
+        ax1.tick_params(axis='y', labelcolor=color)
+        
+        
+        ax2 = ax1.twinx()  # instantiate a second axes that shares the same x-axis
+        color = 'tab:blue'
+        ax2.set_ylabel('Probability', color=color)  # we already handled the x-label with ax1
+        plt.vlines(x=repeated_values, ymin=0, ymax=counts, colors='blue', lw=2, label='vline_single - full height')
+        ax2.tick_params(axis='y', labelcolor=color)
+        plt.show()
```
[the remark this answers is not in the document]
rvare: I'll make it so it returns `fig` and `ax`. However, there is an `ax1` and `ax2`, do both of these need to be returned or only one?

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
rvare: Since I plan on making an if-statement that handles plotting, I'll move that there as I make those changes.

qmcpy/integrand/AmericanOption.py
rvare: For some reason GitHub highlighted the code above the line I've mentioned as well, even though I highlighted the code that was underneath it because it's also related. Not sure why GitHub messed that up.

qmcpy/integrand/LookBackOption.py
rvare: I'll try to delete this once I finish refactoring the all the options to use the abstract option class. I don't want to make too many changes and have everything start breaking because of it. Once this branch gets merged with mine, then I'll remove it.

rvare:
Changed the base branch from `main` to `develop`.

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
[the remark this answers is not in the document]
alegresor: Try to use `fig,ax = pyplot.subplots(ncols=2)` so that `ax` is a `numpy.ndarray` of axes which may be returned

qmcpy/integrand/AmericanOption.py
alegresor: Maybe you don't need the `exercise` variable and can just use `in_money_values` instead. Otherwise use `numpy.copy`

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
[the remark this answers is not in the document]
rvare: I've been dealing with an issue with this possible solution. When doing this, two subplots are made and not one. It seems when you do `ncols=2` two separate plots are made, but they need to be overlaid with each other. I'm still looking into this, but I figure this was something important to note.

qmcpy/stopping_criterion/__init__.py
```diff
@@ -10,4 +10,5 @@
 from .cub_qmc_ml_cont import CubQMCMLCont
 from .cub_qmc_bayes_lattice_g import CubBayesLatticeG, CubQMCBayesLatticeG
 from .cub_qmc_bayes_net_g import CubBayesNetG, CubBayesSobolG, CubQMCBayesNetG, CubQMCBayesSobolG
+from ._cub_qmc_ld_g import _CubQMCLDG
```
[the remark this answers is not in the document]
rvare: I will remove the import statement.

qmcpy/integrand/AmericanOption.py
rvare: Just to reduce the risk of something breaking, I went ahead and used `numpy.copy` so that it would mostly stay the same.

qmcpy/stopping_criterion/.ipynb_checkpoints/_cub_qmc_ld_g-checkpoint.py
aaditj1962161: Looking at it, I myself don't see any changes except the ones Richard recently made. We can probably delete this.

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
```diff
@@ -156,3 +159,36 @@ def set_tolerance(self, abs_tol=None, rel_tol=None):
         """
         if abs_tol != None: self.abs_tol = abs_tol
         if rel_tol != None: self.rel_tol = rel_tol
+
+    def density_estimation(self):
```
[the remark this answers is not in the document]
aaditj1962161: Can you clarify this a bit more? Me and Richard are confused as to what you exactly mean.

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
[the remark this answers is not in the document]
aaditj1962161: How about we do `axes = array([ax1, ax2]) ` and then return axes? If we use the subplots than we'll get two seperate plots and I think twinx() is the best way to have one plot share the x axes and have two seperate y axes?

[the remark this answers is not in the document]
rvare: That makes sense. I originally thought I had to use the object returned from `subplots`, but this makes more sense. I'll see if I can make this change today.

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
```diff
@@ -156,3 +159,36 @@ def set_tolerance(self, abs_tol=None, rel_tol=None):
...
+        approx_solution, data = self.integrate()
```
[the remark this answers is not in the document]
rvare: I'm not sure what class that is nor how that would work. Can you give some clarification about what you had in mind?

qmcpy/stopping_criterion/.ipynb_checkpoints/_cub_qmc_ld_g-checkpoint.py
rvare: I'll go ahead and delete this. However, I'll probably delete the whole folder otherwise will just be tracking an empty folder at that point.

qmcpy/stopping_criterion/Fixed_multilevel.py
aaditj1962161: Guillem doesn't remember about this. So should we just ignore this and not merge it?

qmcpy/integrand/LookBackOption.py
aaditj1962161, line 1: should it not be `return LookBackOption()` instead of `return AmericanOption()`in the spawn function

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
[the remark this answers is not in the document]
alegresor: Start by putting it in `qmcpy/accumulate_data/ld_transform_data.py`, but later we would like it to be general enough to support all stopping criterion

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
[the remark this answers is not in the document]
alegresor: Instead of returning the `kde` object for the continuous part, let the user pass in samples at which to evaluate the full KDE. Be careful as there is both the continuous part and discrete parts of the KDE.

qmcpy/stopping_criterion/Fixed_multilevel.py
alegresor: Sure, that works

qmcpy/integrand/LookBackOption.py
alegresor: Yes, spawn should return a new instance of the class its spawning with the same parameters

qmcpy/stopping_criterion/Fixed_multilevel.py
aaditj1962161: I added the following to `.gitignore`: `stopping_criterion/Fixed_multilevel.py`. Did it work?

qmcpy/integrand/LookBackOption.py
rvare: I can make this change today and push it once it's done.

qmcpy/integrand/LookBackOption.py
rvare: I went a head and fixed this. `LookBackOption` and `AmericanOption` no longer override `f` anymore.

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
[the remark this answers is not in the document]
aaditj1962161: Updated. Pls check

qmcpy/stopping_criterion/Fixed_multilevel.py
aaditj1962161: Deleted the file

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
```diff
@@ -156,3 +158,63 @@ def set_tolerance(self, abs_tol=None, rel_tol=None):
...
+                    densities[i] = kde_discrete
+                    discrete_indices.append(i)
+                    discrete_values.append(values[i])
+            discrete_indices = array(discrete_indices)
+            discrete_values = array(discrete_values)
+            print("The values from KDE:", densities)
+            if len(discrete_indices) > 0 :
+                values = delete(values,discrete_indices)
+                print("The values whose probability was computed using Probability Mass Function (PMF):", discrete_values)
+            print("The values whose probability was computed using Probability Density Function (PDF):", values)
```
alegresor, line 199: Return these values and let them user print them if they choose

qmcpy/stopping_criterion/_cub_qmc_ld_g.py
```diff
@@ -156,3 +158,63 @@ def set_tolerance(self, abs_tol=None, rel_tol=None):
...
+            """
+            values = atleast_1d(values)
+            if values.ndim != 1:
+                raise ParameterError("Values must be a vector of n points.")
+            densities = kde_cont.evaluate(values)*prob
+            discrete_indices = []
+            discrete_values = []
+            for i in range(len(values)):
+                if values[i] in repeated_values:
+                    densities[i] = kde_discrete
```
alegresor, line 190: `densities.ndim==1` yet we are assigning an element to `kde_discrete` which is a vector? I think you need to find the index `j` where `values[i]` is in in `repeated_values` then set `densities[i] = kde_discrete[j]`. Let me know if I'm misunderstanding

alegresor (changes requested):

demos/talk_paper_demos/SURE_2024_Demos/LookBack_Adaptive_Tests.ipynb
aaditj1962161, line 1: Would this need to be updated a bit considering the density function has been updated

rvare: Most likely. I'll take care of updating it which shouldn't take me too long.

PR #356, opened 2024-07-11
Fix readthedocs failure on Linux

sou-cheng-choi:


alegresor (approved):
Hi @schoi32. If the problem persists feel free to merge directly into develop as only develop and master get compiled on ReadTheDocs when we push

PR #362, opened 2024-10-08
QMCToolsCL integration

alegresor:
I've developed a low level OpenCL / C package for generating low discrepancy point-sets called QMCToolsCL at https://qmcsoftware.github.io/QMCToolsCL This PR integrates the capabilities of this package into QMCPy including 
- Parallel capabilities with OpenCL
- IID randomizations (replications) of a LD point-set without spawning
- Nested Uniform Scrambling (NUS) for Halton
- Gray code Lattice ordering 

A large number of files are changed because the same seed now produces updated LD points with the new randomization routines. This requires updated doctest results across almost all objects.


qmcpy/kernel_methods/fast_transforms/ft_qmctoolscl.py
```diff
@@ -0,0 +1,81 @@
...
+    1 dimensional Fast Walsh Hadamard Transform (FWHT) . 
+    Requires the size of the last dimension is a power of 2. 
+
+    Args:
+        x (ndarray): array of samples at which to run FWHT. Requires x.ndim<=3
+    
+    Returns:
+        ndarray: FWHT values
+    """
+    x,shape,d,n,n_half = _parse_ft_input(x)
```
zitterbewegung, line 76: Why do you do a restructuring bind inside a method? Would rather the method have the function signature.
def fwht(x,shape,d,n,n_half): and then perform the parse?

qmcpy/discrete_distribution/lattice/lattice.py
```diff
@@ -132,93 +144,137 @@ def __init__(self, dimension=1, randomize=True, order='natural', seed=None,
...
                 self.d_max = int(parts[-3])
                 self.m_max = int(parts[-2])
             elif isfile(generating_vector):
-                self.gen_vec_og = load(generating_vector).astype(uint64)
+                self.gen_vec_og = load(generating_vector).astype(uint64)[None,:]
                 self.d_max = int(parts[-3])
                 self.m_max = int(parts[-2])
-            elif "LDData"==generating_vector[:6] and repos.exists("https://raw.githubusercontent.com/QMCSoftware/"+generating_vector):
-                datafile = repos.open("https://raw.githubusercontent.com/QMCSoftware/"+generating_vector)
+            elif "LDData"==generating_vector[:6]:
```
zitterbewegung, line 189: Is there any way we can refactor this out into the package? Relying on a download could fail if GitHub is down.

qmcpy/discrete_distribution/iid_std_uniform.py
```diff
@@ -19,7 +19,7 @@ class IIDStdUniform(IID):
         spawn_key       ()
     """
 
-    def __init__(self, dimension=1, seed=None):
+    def __init__(self, dimension=1, replications=1, seed=None):
```
zitterbewegung, line 22: You don't have to do this but I'm nitpicking if I were to add a argument to a constructor I would have the new argument be always last.

qmcpy/kernel_methods/fast_transforms/ft_qmctoolscl.py
alegresor: `fftbr`, `ifftbr`, and `fwht` are high level wrappers around `qmctoolscl.{fft_bro_1d_radix2,ifft_bro_1d_radix2,fwht_1d_radix2}` which have signatures like you suggest. I do a restructuring bind to use the common parser for all three high level wrappers trying to keep an api like `numpy.fft`

qmcpy/discrete_distribution/iid_std_uniform.py
alegresor: Good suggestion, just moved it

qmcpy/discrete_distribution/lattice/lattice.py
alegresor: The default generating vector for lattices and generating matrices for digital nets are distributed with the package and do not require an internet connection to access. I added cases to `lattice.py` and `digital_net_b2.py` that first check for a local path to the file for those who have cloned `QMCSoftware/LDData/` repo to use offline. The case you highlighted will download the file for users if it is not available locally given that they have have internet access and GitHub is up and running.

qmcpy/kernel_methods/fast_transforms/ft_qmctoolscl.py
zitterbewegung: Ok just explain that in the comments

test/fasttests/test_discrete_distribs.py
```diff
@@ -9,36 +9,34 @@
...
-            self.assertEqual(get_unsigned_long_size_cf(),4)
-        else:
-            self.assertEqual(get_unsigned_long_size_cf(),8)
-
-    def test_size_unsigned_long_long(self):
-        get_unsigned_long_long_size_cf = c_lib.get_unsigned_long_long_size
-        get_unsigned_long_long_size_cf.argtypes = []
-        get_unsigned_long_long_size_cf.restype = ctypes.c_uint8
-        self.assertEqual(get_unsigned_long_long_size_cf(),8)
+    # def test_size_unsigned_long(self):
```
zitterbewegung, line 12: Why is this commented out?

alegresor: Not sure, I uncommented it in the latest commit.

zitterbewegung:
Looked through code I don't see any issues and some suggestions were addressed.

PR #372, opened 2025-01-27
Created qmc_apply_style() to allow for qmcpy.mplstye for regular users

zitterbewegung:
Right now we were using the following to get the mplstyle (matplotlib style) to apply that style for notebooks.

`pyplot.style.use('../qmcpy/qmcpy.mplstyle')`

This will not work since it assumes that you have downloaded the git repository and demos directory since the relative directory to load it qmc.mplstyle is in QMCSoftware/qmcpy/ .

This adds the ability to call the following method that will allow you to use it if you load qmc.mplstyle by just installing it doing `pip install qmcpy`

`qmc_apply_style() or qm.qmc_apply_style() depending upon import.
`
[test_evidence.txt](https://github.com/user-attachments/files/18560734/test_evidence.txt)

Notebooks that have been updated.

- vectorized_qmc_bayes.ipynb
- vectorized_qmc.ipynb
- dakota_genz.ipynb


zitterbewegung:
Canceling due to discussion.

zitterbewegung:
After testing the removal of the mplstyle it will break vectorized_bayes so at minimum to run the notebook it requires latex.

sou-cheng-choi:
We no longer need this right by pull request #373?

sou-cheng-choi:
Do we still need this by pull request #373?

zitterbewegung:
> We no longer need this right by pull request #373?

Yes

PR #377, opened 2025-02-10
Remove mplstyle and add colab compatibility , set output of figures to the current working directory and also add missing scikit-learn dependancies.

zitterbewegung:
This fixes the notebooks that have qmcpy.mpstyle in them and also other depenancies and includes @sou-cheng-choi 
s fix for LaTeX. This was tested also on a local install.

Screenshot 2025-02-09 at 19 40 46

On google colab it will install qmcpy and LaTeX since we know it will use Ubuntu. 

Root directory is now set to '.' which will use the current working directory.

!pip install scikit-learn is added to vectorized-qmcpy and vectorized-qmcpy-bayes due to a required dependency for scikit-learn.


sou-cheng-choi:
The notebook dakota_genz.ipynb has an error:

https://github.com/QMCSoftware/QMCSoftware/blob/refs/heads/remove_mplstyle/demos/dakota_genz.ipynb

PieterjanRobbe:
I still have the data file, but it's 800MB (It contains 262144 points in 128 dimensions). You probably don't want it in the repo as is. If you want to generate it on the fly you'd need to rely on a working dakota installation in your CI/CD pipeline, so that's probably not an option either. We could opt to reduce the number of points (`ns`) or the number of dimensions (`ds`) so the data file is a reasonable size and the test passes.

sou-cheng-choi:
Here’s a link to the file (WeTransfer) from Pieterjan --- many thanks!

https://we.tl/t-0X0NiBcpfR

zitterbewegung:
Testing dakota_genz.ipynb 

- [ ] Passed Local test
- [ ] Passed Colab test

zitterbewegung:
> (1) The notebook dakota_genz.ipynb has an error:
> 
> https://github.com/QMCSoftware/QMCSoftware/blob/refs/heads/remove_mplstyle/demos/dakota_genz.ipynb
> 
> (2) Is it possible to hide the long output from LaTeX installation in the first code cell?
> 
> (3) Perhaps add a button "Open in Colab" in each notebook (see MC_vs_QMC.ipynb for example)?

2 and 3 have been addressed.

PR #388, opened 2025-06-14
Abstract option class

rvare:
Among certain integrand classes for stock options, specifically for the Asian Option, European Option, Lookback Option, and Barrier Option, contained a lot of duplicate code and member variables that could be refactored into an abstract class. In this pull request, a new abstract class called `Option` in the `_option.py` file was created to contain all the similar code that these options contained. Below is a UML class diagram (with some things slightly omitted for simplicity's sake) to illustrate how this abstract class is used by the other option classes.

```mermaid
classDiagram

class Integrand

class Option
	Option : -sampler list~int~
	Option : -parameters list~str~
	Option : -volatility float
	Option : -start_price float
	Option : -strike_price float
	Option : -interest_rate float
	Option : -t_final float
	Option : -true_measure BrownianMotion
	Option : -call_put str
	Option : -multilevel_dims list~int~
	Option : -max_level int
	Option : -leveltype str
	Option : -parent boolean
	Option : parameters list~str~
	Option : _get_discounted_payoff()
	Option : _dimension_at_level() List~int~

class EuropeanOption
	EuropeanOption : +get_exact_value()

class AsianOption

class AmericanOption

class LookbackOption

class BarrierOption

Integrand <|-- Option

Option <|-- AsianOption
Option <|-- EuropeanOption
Option <|-- AmericanOption
Option <|-- LookbackOption
Option <|-- BarrierOption
```

I made some tests in a Jupyter notebook called `AbstractOptionTests.ipynb`. I assume the tests I made should be correct, but please do check.

I was unable to make the American Option work, but as discussed, it's more important to get this abstract class done and get it into the `develop` branch.


rvare:
I have not touched the density estimation for a while, and it might be a good idea to do that in a separate PR, that way we don't accidentally merge in work that's not finished. To be honest, I think what happen was when I created the branch to get both done, but I guess I forgot and should've done things in their own branches, that's my fault as I should've thought more about that.

Since you're doing a major overhaul of qmcpy in `distrib_doc`, it might be a good idea to do like you said and do a PR into that branch and merge the two option classes. Or I could make a new branch from `distrib_doc` or from the branch that `distrib_doc` came from and copy the code I've done so far into that to avoid merging the density estimation with everything else and possibly avoid merge conflicts. We can do that or wait until our next meeting to discuss it more.

alegresor:
Make a new branch off `distrib_docs` with you work and then PR back into `distrib_docs`. Please update the `FinancialOption` class in `distrib_docs` if there is anything in your branch that is not yet included there. Also, the KDE stuff can go in the PR

PR #419, opened 2025-11-21
Parsl presentation from Parslfest 2025

zitterbewegung:
In this pull request these are the slides for Parslfest 2025 in the TalkPaperDemos folder that is part of the demos.


zitterbewegung:
@sou-cheng-choi please review.

sou-cheng-choi:
Thank you for the update.

I appreciate you offering to handle the history removal. To ensure we do not accidentally lose anything important, let us prioritize caution. 

1. Make the requested changes as you have planned
2. Before proceeding with any history removal, please provide the exact commands you plan to use.  This will allow us to understand the process and identify any potential risks.
3. Fix the CI test failures

.gitattributes
```diff
@@ -0,0 +1,2 @@
+demos/talk_paper_demos/Parslfest_2025/Parsl_Testbook_Speedup.deck filter=lfs diff=lfs merge=lfs -text
+demos/talk_paper_demos/Parslfest_2025/Parsl_Testbook_Speedup.pptx filter=lfs diff=lfs merge=lfs -text
```
alegresor: I don't see these files in the branch.

alegresor (changes requested):
See minor comments below. I don't think the deck or slides got uploaded. I would also recommend PDF format rather than PPTX.

.gitattributes
zitterbewegung: These are stored in git-lfs did you do  

`git lfs fetch --all && git lfs checkout  --all `

after cloning the repo?

alegresor (approved):

PR #428, opened 2025-12-05
Move demos/parsl_fest_2025/* to demos/ParslFest2025/

sou-cheng-choi:


test/booktests/READEME.md
zitterbewegung: Rename the file to README.md

PR #460, opened 2026-01-07
Add Paper folder with first paper submitted to JOSS

sou-cheng-choi:


alegresor:
Let's release QMCPy v2.2 then we can finish this paper, hopefully by the end of next week, Jan 16.

Issue #387, opened 2025-06-10
Add Resume Feature for QMCPy Integrators


sou-cheng-choi:
It would be nice to add to QMCPy a "resume" feature for all major stopping criteria. This allows users to:

- Start an integration with a loose tolerance.
- Optionally save the integration state to disk.
- Resume later with a tighter tolerance or in a new session, reusing all previous samples and avoiding redundant computation.

This is a great idea.

sou-cheng-choi:
Handled in PR508

PR #508, opened 2026-04-24
Resume Feature in Stopping Criteria and Iteration Log

sou-cheng-choi:
Details in https://github.com/QMCSoftware/QMCSoftware/issues/387

Plus an iteration log for solvers.


alegresor (changes requested):
Very significant update @sou-cheng-choi! A few comments 

1. We discussed the ML(Q)MC Continuation algorithms not being compatible with resuming since they use a decreasing sequence of tolerances within the algorithm. I see you updated these classes to implement the resume feature. Were you able to resolve the issue we discussed? Are the results from the original implementations reproducible? 
2. The `long demo` linked in your `Resume Feature` demo is not working when I render the docs locally 
3.  In the "Iteration Log" blog the link to `MCQMC2022_Article_Figures.ipynb` is not working correctly when I render the docs locally. 
4. I took a look at the resume demo and both blogs, they look nice! Are there any specific changes to the core algorithms in the `QMCSoftware/qmcpy/` directory I should review closert? There are quite significant changes that I have not been able to follow by reading the file diffs.

sou-cheng-choi:
@alegresor 

Many thanks for your review.  

The links are fixed and should work once this PR is merged.

With regard to your points 1 and 4, the following change in (1) has been made to the ML Cont solvers after our discussion: 

(1) In the first RESUMED step, total sample count, `n_total`,  is efficiently recomputed from per-level sample counts (`data.n_level.sum()`) in CubMLMCCont or `self.replications * data.n_level.sum()` in CubMLQMCCont, to match the value of `n_total` from the last step of the corresponding  LOOSE run.

(2) Subseuqnt sample increments (same as before)

-  In CubMLMCCont, additional per-level samples are computed as the nonnegative difference between the new target and the current count: `np.maximum(0, n_samples - data.n_level[:data.levels + 1])`
-  CubMLQMCCont update path grows only active levels, setting the target count to `n_init` or `2 * data.n_level[l]` for newly generated samples

(3) So if a resumed run and a fresh run terminate with the same final per-level sample-count vector `n_level` (and replications for MLQMC), their final `n_total` values will agree by construction.  (The final number of iterations may not be the same at termination for the two runs.)

alegresor (approved):
Please let me know if there are any specific parts you would like me to review further. Otherwise it looks like a great contribution to me!

sou-cheng-choi:
Thank you and @fjhickernell  for the careful review and continuous oversight. I appreciate the excellent feedback and will proceed with the merge now.

PR #513, opened 2026-05-05
DNB2 speedup

alegresor:
Please do not merge this in yet, @Yifei-Xiong and I are still debugging. This PR just helps us isolate our diffs compared to the `develop` branch

Update the `DigitalNetB2` object with a fused implementation from `qmctoolscl >= 1.2.2` which avoids generating integer representation of base 2 digital net points when only floats are returned.


qmcpy/discrete_distribution/digital_net_b2/digital_net_b2.py
```diff
@@ -630,6 +630,8 @@ def __init__(
             )
 
     def _try_gen_samples_float(self, r, n, d, n_start, mmax, r_x, return_binary):
+        if platform.system() == "Darwin": # MacOS seems to still be faster with the original non-fused implementation for some reason. 
+            return None
         if return_binary or "NUS" in self.randomize:
```
alegresor: Added this @Yifei-Xiong, what do you think? Maybe this is an issue with memory layout like @fredhickernell  mentioned?

Yifei-Xiong: Let me take a look. I think it might be because the innermost loop in the C code isn't running on the last dimension of the array, which could cause cache misses. @alegresor

Yifei-Xiong: @alegresor I have submitted a pull request to the QMCToolsCL, which swapped the order of the loops to improve cache hits. Could you try comparing the performance improvements across different platforms again?

Yifei-Xiong (approved):

Issue #504, opened 2026-04-24
Brownian Bridge implementation


fjhickernell:
Add Brownian bridge to PCA and Cholesky

PR #523, opened 2026-05-18
Add Brownian Bridge path construction as decomp_type option (issue #504)

park1428:
Adds `BrownianBridge` as a third `decomp_type` option in `BrownianMotion`, alongside `PCA` and `Cholesky`. Paths are built with terminal value first, then midpoints recursively. Used the Brownian Bridge conditional mean and variance formulas.


sou-cheng-choi (commented):
Please add unit tests and a demo.

park1428:
I have added the unit tests and demo.

sou-cheng-choi:
We appreciate this nice addition!  

Please consider the following issues. 

1.  Please reinstall the latest qmcpy environment. This requires merging the content from the `develop` branch into your local copy. After the merge, set up the qmcpy environment again using Python version 3.13 following up CONTRIBUTING.md. Finally, run your files for a complete sanity check.
2.  `gaussian.py`:
    a) Update the stale message on line 120: `raise ParameterError("decomp_type should be 'PCA' or 'Cholesky'")`.
    b) When a user executes `Gaussian(DigitalNetB2(4), decomp_type='BrownianBridge')`, the current `NotImplementedError` is confusing. Please change this error type to a `ParameterError`.
3.  `brownian_motion.py`: On line 131, the assignment `z_idx = [1]` is mutable. Replace this with `itertools.count()`.
4.  `geometric_brownian_motion.py`: Update line 67:  `"Method of decomposition, either 'PCA' or 'Cholesky'."`
5.  `mkdocs.yml`: Please add the demo path after line 60.

fjhickernell:
Have we moved to Python 3.13?  CONTRIBUTING.md says 3.12.

Best,
Fred


Fred J. Hickernell
Professor, Department of Applied Mathematics
Center for Interdisciplinary Scientific Computation (CISC) cos.iit.edu/cisc/<https://cos.iit.edu/cisc/>
Illinois Institute of Technology
Galvin Tower 18C3-1, 10 W 35th St, Chicago, IL 60616
***@***.******@***.***>, +1 312 567 8983






On May 20, 2026, at 11:18 AM, sou-cheng-choi ***@***.***> wrote:

[https://avatars.githubusercontent.com/u/18743024?s=20&v=4]sou-cheng-choi left a comment (QMCSoftware/QMCSoftware#523)<https://github.com/QMCSoftware/QMCSoftware/pull/523#issuecomment-4500368908>

We appreciate this nice addition!

Please consider the following issues.

  1.  Please reinstall the latest qmcpy environment. This requires merging the content from the develop branch into your local copy. After the merge, set up the qmcpy environment again using Python version 3.13 following up CONTRIBUTING.md. Finally, run your files for a complete sanity check.
  2.  gaussian.py:
a) Update the stale message on line 120: raise ParameterError("decomp_type should be 'PCA' or 'Cholesky'").
b) When a user executes Gaussian(DigitalNetB2(4), decomp_type='BrownianBridge'), the current NotImplementedError is confusing. Please change this error type to a ParameterError.
  3.  brownian_motion.py: On line 131, the assignment z_idx = [1] is mutable. Replace this with itertools.count().
  4.  geometric_brownian_motion.py: Update line 67: "Method of decomposition, either 'PCA' or 'Cholesky'."
  5.  mkdocs.yml: Please add the demo path after line 60.

alegresor (changes requested):
Thanks for putting this together! A few suggestions are below 

- Include a doctest for Brownian bridge with multiple replications, see the other doctests
- Are there restrictions on the dimension for Brownian bridge? I think it needs d^p-1 for p>0. Is this enforced anywhere? 
- Add a unittest with replications where you manually construct a Brownian bridge for a small dimensional case, e.g., d=7, against your fully automated version. By manual I mean where you index the arrays and concatenate them accordingly for d=7

sou-cheng-choi:
> Have we moved to Python 3.13? CONTRIBUTING.md says 3.12.  

Yes, @fjhickernell . See [CONTRIBUTING.md](https://github.com/QMCSoftware/QMCSoftware/blob/develop/CONTRIBUTING.md)  in develop branch updated this morning.

park1428:
Thank you for all the feedback. 

@sou-cheng-choi, I applied all your suggestions, but I left the error message on line 120 of `gaussian.py` as it is, since adding `BrownianBridge` there would imply it's a valid option for `Gaussian`. Instead, I added eager validation in the `BrownianMotion` constructor that lists all three valid options if the user input is wrong. 

@alegresor, for the dimension check I used a `ParameterWarning` instead of an error, since `BrownianBridge` gives correct results for any number of time steps (it just wouldn't be as efficient), and a hard error would break scenarios like d=52 for weekly option pricing. I also added in another unit test (`test_brownian_bridge_warning_for_non_power_of_2`) that verifies that the warning fires and the result is still accurate.

Please let me know if this is the correct course of action.

sou-cheng-choi:
@park1428 

Thanks for continuing with this algorithm. Meanwhile I have two more questions:

1. When `d` is not a power of 2, the implementation would be more complex; see for example `https://people.maths.ox.ac.uk/gilesm/mc/mc/lec13.pdf`. How does your algorithm would behave or handle cases like `d=7` as @alegresor has inquired?  
2. Do you have a limit for the maximum value of `d` for the algorithm? Could it handle say d=2^2000?

I would appreciate your thoughts.  

Please await @alegresor's re-review before proceeding with the merge.

park1428:
@sou-cheng-choi, thank you for the continual advice!

Currently, the algorithm handles cases like d=7 by doing integer division (`mid = (left + right) // 2`), which means that it can run, but creates an uneven tree. Time steps with lower variance may mistakenly end up receiving higher quality dimensions of the QMC sequence. I think that I could fix this issue by creating a list and sorting it based on highest to lowest variance before assigning the dimensions. I could create and push this version soon, please let me know if you have any objections. 

The algorithm could not handle something like 2^2000, I believe python's recursion limit is 1000 and due to RAM constraints it would likely not get close. Would you like me to create a restriction for this?

qmcpy/true_measure/brownian_motion.py
```diff
@@ -41,6 +45,39 @@ class BrownianMotion(Gaussian):
...
+            mean            [0. 0. 0. 0.]
+            covariance      [[0.25 0.25 0.25 0.25]
+                             [0.25 0.5  0.5  0.5 ]
+                             [0.25 0.5  0.75 0.75]
+                             [0.25 0.5  0.75 1.  ]]
+            decomp_type     BROWNIANBRIDGE
+
+        With Brownian Bridge construction and independent replications
+
+        >>> x = BrownianMotion(DigitalNetB2(4,seed=7,replications=2),decomp_type='BrownianBridge')(4)
```
alegresor: Change to `BrownianMotion(DigitalNetB2(4,seed=7,replications=3),decomp_type='BrownianBridge')(2)` so `x.shape == (3,2,4)`, having shape `(2,4,4)` may confuse viewers on which 4 corresponds to the number of samples and which corresponds to the number of dimensions.

alegresor (changes requested):
Thanks for addressing my previous feedback @park1428! This is quite a nice PR. 

My only remaining concern is the Brownian bridge construction when `d` is not a power of `2`. Could you add a section detailing the math of the Brownian bridge construction to your notebook, and specify what you are doing when `d` is not a power of `2` mathematically? Perhaps add some references as well, this will bring you closer to a potential future blog post! 

I think it would also be helpful to have a figure with say 6 plots where in each plot you increase the dimension, say from 4 to 5 to 6 to 7 to 8 to 128. Keeping the same seeding should yield the same paths, just refined, helping readers understand the iterative construction of Brownian bridge by taking midpoints. It would also help us understand at what time-steps you are adding for non-powers-of-2.

sou-cheng-choi:
@park1428 

I would also be happy if we allow arbitrary $d$ but document and warn that it is not variance-optimized.

Perhaps you could initiate a 30-minute review meeting with @fjhickernell , @alegresor , and me some time this or early next week. It could be really helpful. 

Two more thoughts for now:

(a) bm = BrownianMotion(DigitalNetB2(6, seed=7), decomp_type='BrownianBridge',lazy_decomp=False) throws an error, but it is better to ignore `lazy_decomp` value. 
    
(b) It is probably better to change `decomp_type` to `construction_type` .  The change should however be made backward compatible; see, e.g., https://peps.python.org/pep-0387/

fjhickernell:
Yes, suggest some times that you are available.  Let’s have the four of us take a look where you can walk us through it.

Best,
Fred


Fred J. Hickernell
Professor, Department of Applied Mathematics
Center for Interdisciplinary Scientific Computation (CISC) cos.iit.edu/cisc/<https://cos.iit.edu/cisc/>
Illinois Institute of Technology
Galvin Tower 18C3-1, 10 W 35th St, Chicago, IL 60616
***@***.******@***.***>, +1 312 567 8983






On May 26, 2026, at 2:39 PM, sou-cheng-choi ***@***.***> wrote:

[https://avatars.githubusercontent.com/u/18743024?s=20&v=4]sou-cheng-choi left a comment (QMCSoftware/QMCSoftware#523)<https://github.com/QMCSoftware/QMCSoftware/pull/523#issuecomment-4547993793>

@park1428<https://github.com/park1428>

I would also be happy if we allow arbitrary $d$ but document and warn that it is not variance-optimized.

Perhaps you could initiate a 30-minute review meeting with @fjhickernell<https://github.com/fjhickernell> , @alegresor<https://github.com/alegresor> , and me some time this or early next week. It could be really helpful.

park1428:
Sounds good! I'm available around noon this Friday, or anytime after 4 pm on Tuesday or Wednesday next week.

alegresor:
> Sounds good! I'm available around noon this Friday, or anytime after 4 pm on Tuesday or Wednesday next week.

Any of those times work for me.

sou-cheng-choi:
I prefer this week if possible.

fjhickernell:
Just sent a Zoom calendar invite for Friday noon.

Best,
Fred


Fred J. Hickernell
Professor, Department of Applied Mathematics
Center for Interdisciplinary Scientific Computation (CISC) cos.iit.edu/cisc/<https://cos.iit.edu/cisc/>
Illinois Institute of Technology
Galvin Tower 18C3-1, 10 W 35th St, Chicago, IL 60616
***@***.******@***.***>, +1 312 567 8983






On May 27, 2026, at 10:49 AM, sou-cheng-choi ***@***.***> wrote:

[https://avatars.githubusercontent.com/u/18743024?s=20&v=4]sou-cheng-choi left a comment (QMCSoftware/QMCSoftware#523)<https://github.com/QMCSoftware/QMCSoftware/pull/523#issuecomment-4556186274>

I prefer this week if possible.

alegresor:
@park1428, I made pushed a small changes to your notebook with a quick assertion that doubling the dimensions retains the same samples at the previous time step. I think you need to modify your code to get this working. I also updated the increasing d plot. You might take a look at Section 6.4 of Art Owen's book https://artowen.su.domains/mc/, especially figure 6.9

park1428:
@alegresor, I believe the reason the assertion fails may be that even with seed=7, the sequences for d=2 and d=4 aren't the same in their shared dimensions. I tried printing the sequences and the values do not match, and I found that using randomize='DS' makes the dimensions match. 
```python
seq_d2 = DigitalNetB2(2, seed=7)(1)
seq_d4 = DigitalNetB2(4, seed=7)(1)
print(seq_d2)  # [[0.72162356 0.914955  ]]
print(seq_d4)  # [[0.62154373 0.69551885 0.07905092 0.26156152]]
```
What direction do you think I should go with this?

alegresor:
Good catch @park1428. Switch over to using `qp.Lattice` instead of `qp.DigitalNetB2`. This should satisfy the property and help us visualize the paths filling in with increasing dimension.

sou-cheng-choi:
@park1428 

1. In the latest version, code cell [7] of brownian_bridge.ipynb is broken. As a result, CI tests are also broken. Please fix the issue.
2. Please make sure `bm = BrownianMotion(DigitalNetB2(6, seed=7), decomp_type="BrownianBridge", lazy_decomp=False)` will proceed with the calculations. Currently, it is throwing an error.
3. Please preserve states in the following example:

```
bm = BrownianMotion(DigitalNetB2(4, seed=7),  initial_value=2, diffusion=4, decomp_type="BrownianBridge")
spawned = bm._spawn(DigitalNetB2(4, seed=7), 4)
print(bm.initial_value, bm.diffusion)                       # 2  4  (correct)
print(spawned.initial_value, spawned.diffusion)   # 0    1    (bug: state was dropped)
```

Please add fast unit tests to ensure the last two bugs won't occur again in the future.

sou-cheng-choi:
@park1428 @fjhickernell @alegresor 

A Slack channel "#brownian-bridge-josh-park" has been created.

fjhickernell:
Please let us know when you are ready for us to give this another code review.

sou-cheng-choi:
> Please let us know when you are ready for us to give this another code review.

@park1428 

You could request for re-review in this PR whenever you are ready. Happy Independence Day weekend!

park1428:
@alegresor, I switched to using DigitalNetB2 to create the van der Corput sequence and found that it's a little slower than the original method for lower dimensions (1.5x at d=4). It starts to become similar in speed at higher dimensions (at d=256 and d=1024 the two methods are roughly the same). I wanted to know, what do you think about this?

alegresor:
> @alegresor, I switched to using DigitalNetB2 to create the van der Corput sequence and found that it's a little slower than the original method for lower dimensions (1.5x at d=4). It starts to become similar in speed at higher dimensions (at d=256 and d=1024 the two methods are roughly the same). I wanted to know, what do you think about this?

That's fine, it is probably not the biggest contributor to the Brownian bridge construction time anyways. 

On a separate note, I am curious how your VDC implementation beat the DigitalNetB2 construction as DigitalNetB2 should be as fast as possible, especially since it uses Gray code ordering with only one bit-flip per iteration and is implemented in C. Separate from this PR, if you are interested in looking more into a timing comparison between your VDC implementation and the DigitalNetB2 method that could be interesting!

sou-cheng-choi:
@park1428 

The All Tests failures can be resolved by merging `develop` branch into your branch.

sou-cheng-choi (changes requested):

park1428:
I have pushed the final changes from the last meeting (switch to digital net b2, some renaming, "bridge" as a valid input, demo improvements, and the new bridge output parameter).

fjhickernell:
I am only seeing  version from July 6.  Please let me know the commit number.

park1428:
The latest commit is "fix typos" (f3f3f24) and should contain all the updated code

sou-cheng-choi (changes requested):
@park1428 

1. Align error messages for `invalid decomp_type` when `lazy_decomp=False` is passed, ensuring the 'BrownianBridge' type is listed using an example in a unit test like: `BrownianMotion(DigitalNetB2(4, seed=7), decomp_type="invalid", lazy_decomp=False)`. Right now, the error message is `raise ParameterError("decomp_type should be 'PCA' or 'Cholesky'")`
2. Add a check or docstring caveat to prevent users from setting `monitoring_times` that exceed `t_final` in an unit test with an example like `BrownianMotion(DigitalNetB2(4, seed=7), t_final=1.0, decomp_type="BrownianBridge", monitoring_times=[0.1, 0.2, 0.3, 5.0])`.

park1428:
The decomp_type check now runs before the Gaussian setup so that all three types are mentioned even with `lazy_decomp=False`, and having a `monitoring_times` that exceeds `t_final` will raise an error. 

@sou-cheng-choi @alegresor @fjhickernell, ready for another review whenever convenient.

qmcpy/true_measure/brownian_motion.py
```diff
@@ -41,6 +44,72 @@ class BrownianMotion(Gaussian):
...
+               [-0.44240726, -1.34558221, -1.22522271, -1.58454187]])
+
+        With custom output order
+
+        >>> true_measure = BrownianMotion(DigitalNetB2(4,seed=7),decomp_type='BrownianBridge',monitoring_times=[0.6,1.0,0.3,0.8],bridge_output_order='input')
+        >>> true_measure.time_vec
+        array([0.3, 0.6, 0.8, 1. ])
+        >>> true_measure(2)
+        array([[ 0.4363325 ,  0.3095377 , -0.02913874, -0.07341545],
+               [-1.34558221, -1.58454187, -0.44240726, -1.22522271]])
```
alegresor: For each of these doctests, also print the `_construction_times` and the output times (not sure what the attribute name is for this).

If my understanding is correct, `time_vec` is always in increasing order, `_construction_times` is the order in which the BM bridge is constructed, and there should be an attribute for for the time vector in the output order.

alegresor (changes requested):
Looking good @park1428! A few more minor comments. Nothing technical, just clarity suggestions. Looking forward to getting this merged in soon! 

-  In the notebook, your timing comparison is comparing initialization times and generation times. Either rename the axis titles or rework the timing logic to match the current axis titles. 
- Rename `self._construction_times` to `self.bridge_construction_times` and (if it doesn't already exist by another name) add an attribute `self.bridge_output_times` for the times at which the output is returned (see also my line-highlighted comment)
- For the bridge construction, add both these attributes to `self.parameters` so they get printed when the BM object is printed

alegresor:
Please re-request a review from @sou-cheng-choi and myself when you are ready

sou-cheng-choi (changes requested):
@park1428 

Thanks for your continuous efforts into enhancing this feature. We are just tidying some small loose ends now. 

1. It does not make sense to have NaN for the time like in  `BrownianMotion(DigitalNetB2(4, seed=7), t_final=1.0, decomp_type="BrownianBridge", monitoring_times=[0.1, 0.2, np.nan, 1.0])`. 
2. Similarly, `t_final` should not be negative or NaN.  Problematic examples include: `BrownianMotion(DigitalNetB2(4, seed=7), t_final=-8, decomp_type="BrownianBridge")` or `BrownianMotion(DigitalNetB2(4, seed=7), t_final=np.nan, decomp_type="BrownianBridge")`
3. The `Examples` section in the docstring of `brownian_motion,py` seems to be rather long. Could we add labels like Example 1, Examplel 2, etc.?
4. `vdc_ordering` is mentioned in `brownian_bridge.ipynb`, but I can't find it in any Python source files. 
5. `qmcpy/integrand/financial_option.py`'s `decomp_type` mentioned `PCA` and `Choleskey`. We need to add `BrownianBridge` now.
6. `np.concat` is introduced only in NumPy 2.0 (see `https://numpy.org/devdocs/reference/generated/numpy.concat.html`). For earlier versions that QMCPy permits (Numpy 1.17.0 or later), we use `np.concatenate`.

sou-cheng-choi (changes requested):
See above comments

park1428:
@sou-cheng-choi @alegresor, I have addressed all of the issues. 
For the timing graphs, I may be misunderstanding the comment, but currently the first graph represents the initialization + generation while the second represents generation only. I have clarified the titles to be "Initialization and generation of paths" and "Generation of paths only", please let me know if you had something else in mind. 
The latest commit is 2f135d0

alegresor:
> For the timing graphs, I may be misunderstanding the comment, but currently the first graph represents the initialization + generation while the second represents generation only. I have clarified the titles to be "Initialization and generation of paths" and "Generation of paths only", please let me know if you had something else in mind. 

You are correct. Sorry, I misread the code.

alegresor (approved):
Thanks for all the hard work @park1428!

fjhickernell:
@park1428 , @sou-cheng-choi where are we with this?

fjhickernell (approved):
Looks good.

park1428:
@fjhickernell currently all the changes are completed. I unfortunately can't make it to the Tuesday meeting due to classes, if there is anything that should be done on my part for the release before then please let me know. 

@sou-cheng-choi whenever you get a chance, could you please take another look?

sou-cheng-choi (changes requested):
@park1428 

With `bridge_output_order="input"`, generated paths follow the input `monitoring_times` order, but mean, covariance, and PDF remain in increasing-time order.  This can be confusing.  Could you check?

park1428:
@sou-cheng-choi thank you for the review. I have now also reordered those values.

sou-cheng-choi (approved):

fjhickernell:
Joshua looks like this is ready. You can have the privilege of squashing and merging.

PR #541, opened 2026-06-15
Update zero-inflated exponential and add ProductMeasure support for independent true measures

Laasya-73:
### Summary

Update the zero-inflated exponential work so it fits naturally into the QMCPy `TrueMeasure` structure, and extend this issue to include a `ProductMeasure` for combining independent true measures.

The zero-inflated exponential distribution has a discrete point mass at zero and a continuous exponential component for positive values. Let $U \sim \mathrm{Uniform}(0,1)$, with zero probability $p_0 \in (0,1)$ and rate $\lambda > 0$. Define

$$
X =
\begin{cases}
0, & U \le p_0, \\
-\dfrac{1}{\lambda}\log\left(1-\dfrac{U-p_0}{1-p_0}\right), & U > p_0.
\end{cases}
$$

Then

$$
\mathbb{P}(X=0)=p_0,
$$

and for $x>0$, the continuous part has density

$$
f_X(x) = (1-p_0)\lambda e^{-\lambda x}.
$$

### ProductMeasure motivation

The broader goal is to support independent products of existing QMCPy true measures. Suppose we have child true measures $T_1,\ldots,T_k$ with dimensions $d_1,\ldots,d_k$. The product measure has total dimension

$$
d = \sum_{j=1}^{k} d_j.
$$

Given a unit-cube point split into coordinate blocks,

$$
u = \left(u^{(1)},u^{(2)},\ldots,u^{(k)}\right),
$$

where $u^{(j)} \in [0,1]^{d_j}$, the product transform should apply each child transform independently:

$$
T(u)=
\left(
T_1\left(u^{(1)}\right),
T_2\left(u^{(2)}\right),
\ldots,
T_k\left(u^{(k)}\right)
\right).
$$

The transformed outputs are concatenated along the final coordinate axis. Since the components are independent, the combined weight should be

$$
w(x)=
\prod_{j=1}^{k} w_j\left(x^{(j)}\right),
$$

where $x^{(j)}$ is the coordinate block corresponding to the output of $T_j$.

### Proposed work

* Keep the zero-inflated exponential as a QMCPy-compatible true measure.
* Preserve backward compatibility for the old `y_split` argument using a deprecation warning.
* Add a `ProductMeasure` true measure that accepts a list of independent child true measures.
* Split input QMC coordinates according to the child dimensions.
* Apply each child transform independently.
* Concatenate transformed outputs along the final coordinate axis.
* Multiply child weights because the components are independent.
* Add tests for examples such as:

  * zero-inflated exponential plus uniform,
  * Gaussian plus `SciPyWrapper`,
  * child true measures with different dimensions.

### Motivation

This will allow users to build independent mixed measures from reusable QMCPy true measure components instead of writing every combined distribution as one custom joint transform.


sou-cheng-choi (changes requested):
1. Change of public API such as removal of `y_split` could result in run time errors for calls with the input parameter. Consider adding deprecation warning.
2. Please fix the warning from `make doc`: `WARNING -  [mkdocs-print-site] 'demos/scipywrapper_dependence_custom/scipywrapper_demo.ipynb' file is missing a leading h1 tag.`

Laasya-73:
@alegresor, @sou-cheng-choi, @fjhickernell 

I reran the CI, but the `macOS` job is consistently failing during the `Install MPMC dependencies` step, before any of the tests are executed. The log shows it is invoking `Python 3.14` `(/Library/Frameworks/Python.framework/Versions/3.14/bin/python)` even though the workflow is labeled `Python 3.13`, and `pyg_lib` fails to install for the current torch build. As a result, the remaining matrix jobs are cancelled. It looks like the fix would need to be in the CI workflow or dependency setup. I'm not sure if I'm expected to modify the workflow, so I wanted to check with you first.

Thank you:))

sou-cheng-choi:
@Laasya-73 

The failures are being fixed in [PR555](https://github.com/QMCSoftware/QMCSoftware/pull/555). You could wait until it is reviewed and merged into `develop` branch and then merge `develop` into your branch.

alegresor (changes requested):
All the structure looks good to me! A few minor suggestions 

- add doctests for both the product measure and zero inflated exponential. Include both with and without multiple replications. 
- Should the notebook `NonUniformSampler_MWE` be added to the docs? Should the two new notebooks be combined?

fjhickernell:
What is the status?  Laasya, you and I talked about having a dummy measure when you are setting up the children, but I do not see that yet in the product_measure.ipynb demo.  Is that still work in progress, or did I miss something?

Laasya-73:
> What is the status? Laasya, you and I talked about having a dummy measure when you are setting up the children, but I do not see that yet in the product_measure.ipynb demo. Is that still work in progress, or did I miss something?

Yes, this is still work in progress. I’m working on adding the dummy measure setup in the ProductMeasure demo and will push the update shortly. I’ll also address @alegresor 's doctest/notebook suggestions in the same update.

sou-cheng-choi (changes requested):
@Laasya-73 

1. The notebook `product_measure.ipynb` does not have any output cells. 
2.  Made two commits to add the genrated booktest file for the demo and documentation in `docs/api/true_measures.md`.

sou-cheng-choi:
@Laasya-73 

We are seeing the warnings in [unit tests](https://github.com/QMCSoftware/QMCSoftware/actions/runs/29538098883/job/87754274101). Kindly investigate.

```

=============================== warnings summary ===============================
test/test_product_measure.py::test_product_measure_zero_inflated_with_scipy_uniform_shape
test/test_product_measure.py::test_product_measure_replication_shape
test/test_product_measure.py::test_product_measure_marginals_with_different_dimensions
test/test_scipy_wrapper_custom.py::test_zero_inflated_zero_rate
test/test_scipy_wrapper_custom.py::test_zero_inflated_y_split_deprecation
  /home/runner/work/QMCSoftware/QMCSoftware/qmcpy/true_measure/scipy_wrapper.py:341: UserWarning: Custom univariate distribution has no 'pdf' or 'logpdf'. Weights will be treated as 1 for this marginal.
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html

```

Laasya-73:
> @Laasya-73
> 
> We are seeing the warnings in [unit tests](https://github.com/QMCSoftware/QMCSoftware/actions/runs/29538098883/job/87754274101). Kindly investigate.
> 
> ```
> 
> =============================== warnings summary ===============================
> test/test_product_measure.py::test_product_measure_zero_inflated_with_scipy_uniform_shape
> test/test_product_measure.py::test_product_measure_replication_shape
> test/test_product_measure.py::test_product_measure_marginals_with_different_dimensions
> test/test_scipy_wrapper_custom.py::test_zero_inflated_zero_rate
> test/test_scipy_wrapper_custom.py::test_zero_inflated_y_split_deprecation
>   /home/runner/work/QMCSoftware/QMCSoftware/qmcpy/true_measure/scipy_wrapper.py:341: UserWarning: Custom univariate distribution has no 'pdf' or 'logpdf'. Weights will be treated as 1 for this marginal.
>     warnings.warn(
> 
> -- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
> ```

@sou-cheng-choi,

Thanks for checking. These warnings are expected because ```ZeroInflatedExpUniform``` intentionally implements only the inverse CDF (```ppf```) and does not provide a ```pdf``` or ```logpdf```. ```SciPyWrapper``` therefore warns that it treats the weight as 1 for that marginal, which is the intended behavior. Would you prefer that I explicitly capture these warnings in the unit tests, or is it fine to leave them visible since they are expected?

sou-cheng-choi:
@Laasya-73 

1. What about removing the construction-time warning at line 341 in `scipy_wrapper.py` and keep the one-time warning at line 497? 
2. Add a unit test if applicable. 
3. Remove `warnings.simplefilter("ignore", UserWarning)` in doctest in `qmcpy/true_measure/zero_inflated_exp_uniform.py`

qmcpy/true_measure/zero_inflated_exp_uniform.py
```diff
@@ -1,53 +1,117 @@
...
+    >>> tm = ZeroInflatedExpUniform(
+    ...     DigitalNetB2(1, seed=7, replications=2),
+    ...     p_zero=0.4,
+    ...     lam=1.5,
+    ... )
+    >>> x = tm(8)
+    >>> x.shape
+    (2, 8, 1)
+    >>> bool((x >= 0).all())
+    True
```
alegresor: Print the samples as well. This will help flag if there are code updates in the future which change the output values but not the output shape.

alegresor (commented):
Looks good @Laasya-73! Good idea for the `DummySampler`, and I like the demonstrated equivalence between `ScipyWrapper` and `ProductMeausure` in the notebook

alegresor (approved):

Laasya-73:
@fredhickernell @sou-cheng-choi 

Could you please take a look when you get a chance?
Thank You:))

fjhickernell (approved):
Looks good to me.  Thanks Laasya.

sou-cheng-choi:
@Laasya-73 

Apology for this late response. I really appreciate all the enhancements you have made. 

Here are a few questions and comments since my last review. Please add fast unit tests wherever applicable.

1. Since `DummySampler` returns meaningless sampler values,  it is probably better to make sampling raise an explicit “placeholder cannot generate samples” exception. Right now, e.g., `dummy.gen_samples(1)` still returns `array([[0.]])`, whereas `dummy._gen_samples(1,2,False,False)` returns `array([[[-5.92303701e-289]]])`.

2. Many of the import statements in QMCPy can be simplified. For example, `from qmcpy.true_measure import Gaussian` can be changed to `from qmcpy import Gaussian`.

3. `ZeroInflatedExpUniform` has a broken backward-compatibility claim. For example, `ZeroInflatedExpUniform(DigitalNetB2(2), y_split=0.5)` accepts the deprecated `y_split` keyword but raises a DimensionError if the sampler dimension is not 1, failing before issuing any deprecation warning. A valid deprecation requires behavioral preservation (e.g., 2D sampler) alongside warning.

4. Shouldn't the following example fail? 
      
      ```
      >>> ProductMeasure(DigitalNetB2(2), [AcceptanceRejection(DigitalNetB2(2), lambda x: np.ones(len(x)), 1, 1)])
      ProductMeasure (AbstractTrueMeasure)
          marginals       AcceptanceRejection (AbstractTrueMeasure)
                              target_dim      1
                              upper_bound     1
                              density_integral 1
                              acceptance_rate 1
      ```

fjhickernell:
Thanks Sou-Cheng.  Laasya, please address these. We are close.

sou-cheng-choi (changes requested):
Please see my last comment. Could you also review the issue raised by **github-code quality** above?

Laasya-73:
> Please see my last comment. Could you also review the issue raised by **github-code quality** above?

@sou-cheng-choi,

I just pushed an update addressing the latest comments. I also reviewed the GitHub code-quality warning and updated the ZeroInflatedExpUniform initialization path so it does not manually call multiple initializers. Please let me know if anything still looks off.

Thank You:)

sou-cheng-choi:
I approve the changes.

sou-cheng-choi (approved):

PR #548, opened 2026-07-02
Enhance LaTeX install and add a markdown-unwrap tool

sou-cheng-choi:


.github/workflows/alltests.yml
```diff
@@ -270,48 +288,44 @@ jobs:
           eval "$(/usr/libexec/path_helper)"
           sudo tlmgr update --self
           sudo tlmgr install latexmk dvipng collection-fontsrecommended type1cm
```
alegresor: This also seems to fail occasionally

alegresor (changes requested):
The MacOS LaTeX install case also fails occasionally. Gemini suggested the following to me

```yml
     # 1. Install LaTeX on Linux (Ubuntu)
      - name: Install LaTeX on Linux
        if: matrix.os == 'ubuntu-latest'
        run: |
          sudo apt-get update
          sudo apt-get install -y dvipng texlive-latex-extra texlive-fonts-recommended cm-super

      # 2. Install LaTeX on macOS (using Homebrew)
      - name: Install LaTeX on macOS
        if: matrix.os == 'macos-latest'
        run: |
          brew install --cask mactex-no-gui
          echo "/Library/TeX/texbin" >> $GITHUB_PATH

      # 3. Install LaTeX on Windows (using MiKTeX)
      - name: Install LaTeX on Windows
        if: matrix.os == 'windows-latest'
        run: |
          choco install miktex --no-progress
          echo "C:\Program Files\MiKTeX\miktex\bin\x64" >> $GITHUB_PATH
          # Note: MiKTeX will download missing packages automatically on first use
```

sou-cheng-choi:
@alegresor 

Thanks for flagging out the issues that I have overlooked for MacOS.  

I have implemented retry of up to 3 attempts to make the LaTeX installation more robust for both Windows and MacOS. Hopefully it will reduce the frequency of failure.

alegresor (approved):

PR #552, opened 2026-07-04
Fix pep8 workflow error and unskip dockets in MPMC

sou-cheng-choi:
@nmkirk and @alegresor 

I have unskipped doctest in `mpmc.py`. Please take a look of the doctest output and see if the changes are okay.


.github/workflows/docs.yml
```diff
@@ -30,7 +32,7 @@ jobs:
           restore-keys: |
             mkdocs-material-
       - uses: ts-graphviz/setup-graphviz@v2
-      - run: pip install -e .[docs]
```
alegresor, line 33: Why do we need to install all .[dev_no_mpmc] rather than just .[docs]?

.github/workflows/pep8.yml
```diff
@@ -27,7 +27,7 @@ jobs:
 
       - name: Install project with lint dependencies
         run: |
-          python -m pip install -e ".[docs]"
```
alegresor, line 30: Why was this command changed to `python -m pip install -e . "pylint>=4.0.5"`?

alegresor (changes requested):
See questions below

sou-cheng-choi:
@alegresor 

Thanks for your review and questions.

1. I have reversed `dev_no_mpmc` to `docs` in `pyproject.toml` and `docs.yml`.
2. The command python -m pip install -e . "pylint>=4.0.5" in the PEP8 workflow installs the current repository as an editable package (-e .) and ensures pylint is present (version 4.0.5 or higher).
This setup for the PEP8 job is sufficient because make check_pep8 only runs pylint qmcpy, making broader documentation dependencies unnecessary.

alegresor (approved):

PR #554, opened 2026-07-06
JCAM  > JSC (Journal of Scientific Computation)

alegresor:
rename for journal submission


sou-cheng-choi:
@alegresor 

I have made a small commit to fix typo in `docs/booktests.md` and approved the code changes.

In the future, please do not use special characters like `>' in branch names because they would require quotes to work with in comands.

sou-cheng-choi (approved):

PR #555, opened 2026-07-09
Fix recent full sweep test failures

sou-cheng-choi:
The recent release of PyTorch 2.14 on July 7 has resulted in errors due to non-existence of some compatible packages as specified in `pyproject.toml`.


alegresor (approved):
Looks fine to me. My only suggestions are to make @nmkirk aware of this MPMC torch limitation and to keep an eye on it to see if it gets patched so we may remove the `torch < 2.13` requirement from the `mpmc` target in `pyproject.toml` in the future.

Issue #538, opened 2026-06-12
Add mean, standard deviation, variance/covariance to the TrueMeasures


ishn29:

alegresor:
Some of the IID stopping criterion such as `CubMCCLT` and `CubMCG` already set this for their `Data` objects. For low-discrepancy stopping criterion the mean makes sense, but I'm not sure what the interpretation of standard deviation, variance, and covariance would be since the samples are not independent. Regardless, all `Data` objects should have a `yfull` attribute which can be used to early get these metrics.

fjhickernell:
My suggestion about this was not that there be this info in the accumulate_data class, but that these be properties of the distributions that we are trying to emulate.

alegresor:
That makes sense and sounds like a great suggestion to me! 

My apologies for not reading close enough, please disregard my previous message.

PR #558, opened 2026-07-12
mean, standard deviation, variance, and covariance for Uniform, Gaussian, and Kumaraswamy

ishn29:
## Summary

- Issue: https://github.com/QMCSoftware/QMCSoftware/issues/538

This pull request adds mean() and var() methods to the Uniform true measure.

The methods return the component wise mean and variance of the multivariate uniform distribution:

mean() returns (a + b) / 2
var() returns (b - a)**2 / 12


alegresor:
A few ideas 

- Do we want the full `d x d`  covariance matrix or just the length-`d` vector of variances? 
- Maybe implement this for other true measures in this PR as well. 
- Instead of having mean and variance functions, make these attributes with common names across measures, e.g., `self.mean` and `self.var` or `self.cov`. Try to be consistent with the `Gaussian` object which already has attributes.

ishn29:
Thanks for the suggestions! I updated the implementation to follow existing `Gaussian` object by replacing the `mean()`/`var()` methods with `self.mean` and `self.covariance` attributes.

One question I had is whether `mean` and `covariance` should also be added to `self.parameters`, or whether `parameters` should only have constructor arguments since these are derived from the upper and lower bounds.

I'll also extend this to the other applicable true measures in this PR.

alegresor:
> One question I had is whether `mean` and `covariance` should also be added to `self.parameters`, or whether `parameters` should only have constructor arguments since these are derived from the upper and lower bounds.

Yes, please add to `self.parameters` as well, good idea! You'll need to update quite a few doctests with this change, but that is fine!

Please let me know when you are happy with the PR and I'll give it a review.

sou-cheng-choi:
@ishn29 All Tests are failing.  Try to merge branch `develop` into your branch. It might help.

ishn29:
- Added `mean`, `variance`, `standard_deviation`, and `covariance` attributes to the following true measures:
  - `Uniform`
  - `Kumaraswamy`
  - `Gaussian`
- Added the new attributes to each true measure’s `parameters` list so they appear in printed representations and doctests.
- Standardized shapes of the attributes:
  - `mean`, `variance`, and `standard_deviation` are length `d` arrays.
  - `covariance` is a `d × d` matrix.
- Updated the `MaternGP` and `Lebesgue` doctests to include the expanded Gaussian and Uniform statistics.
- Updated integrand and stopping criterion doctests wherever their displayed true measures now include the new statistical attributes.
- Added `demos/statistics_for_TrueMeasure.ipynb`, which:
  - Demonstrates the new true measure statistics.
  - Compares theoretical statistics with sample estimates.
  - Includes examples for Uniform, Kumaraswamy, and Gaussian measures.

alegresor (changes requested):
Looks good @ishn29! A few suggestions are below.

-  In the notebook, when you print values in the table use scientific notation so it doesn't show a bunch of zeros for values below 5e-4
- In the notebook, you have some nice convergence tests for Uniform, but less thorough for Gaussian and Kumaraswamy. Perhaps you could add one more level of abstraction to your uniform code blocks which loops over the other two distributions as well, giving tables and plots for all three examples. 
- Where did you find the formulas for the Kumaraswamy mean and variance? Add a citation into the docstring for that object. 
- Add some unit tests which numerically verify the mean and covariance matrix for the objects. Say draw 2^15 samples from the object and then verify the sample mean and sample cov are within 1e-5 everywhere. This is a watered-down version of what you have in the notebook, but it will quickly flag any errors if code gets updated in the future which accidentally changes sample means or covs.

sou-cheng-choi (changes requested):
@ishn29 

I have the following suggestions. For items 3 - 5, please add unit tests:

1. You have yet to list your demo notebook in `mkdocs.yml`.
2. I have added a unit test file `test/booktests/tb_statistics_for_TrueMeasure.py` for your demo. 
3. Fix the following error.
    ```
    matern = MaternGP(IIDStdUniform(3, seed=7), points=np.linspace(0, 1, 3)[:, None], variance=0.01)
    matern._spawn(IIDStdUniform(3, seed=8))
    ```
4. Fix the following problematic computed values:
   ```
    >>> k = Kumaraswamy(IIDStdUniform(1, seed=7), a=42169.650342858, b=562341.32519034)
    >>> print(f"{k.variance=}, {k.standard_deviation=}")
     k.variance=array([-1.47344914e-09]), k.standard_deviation=array([nan])
    ```
 5.  In the following example, upper bound is less than lower bound, and returned negative standard deviation:
     ```
      >>> u = Uniform(IIDStdUniform(1, seed=7), lower_bound=[1], upper_bound=[0])
      >>> print(f"{u.variance=}, {u.standard_deviation=}")
      u.variance=array([0.08333333]), u.standard_deviation=array([-0.28867513])
      ```

fjhickernell:
I would actually keep the standard deviation and variance, even though they are redundant.  They are convenient.

alegresor:
> I would actually keep the standard deviation and variance, even though they are redundant. They are convenient.

Makes sense, let's keep them! I edited my old list of requested changes accordingly.

sou-cheng-choi (changes requested):
@ishn29 

1. Variance and standard deviation is missing for BrownianMotion instances.
2. `self.variance = second_moment - self.mean**2` in `qmcpy/true_measure/kumaraswamy.py` results in _cancellation error_ when subtraction of two numbers in similar magnitude are involved. So the Kumaraswamy example in my last review stil returns invalid statistics. 
3.  `Uniform(IIDStdUniform(1, seed=7), lower_bound=np.nan)` accepts NaN as input value. Please validate input values.
4. For  covariance matrices of `Uniform(IIDStdUniform(d))` and `Kumaraswamy(IIDStdUniform(d))` that have zeros in off-diagonal entries, we just need to store the diagonal vector to save memory consumption from $O(d^2)$ to $O(d)$.
5. We should ensure read-only views for statistical parameters, including mean, variance, and covariance.
This can be achieved by utilizing private attributes (e.g., self._mean) combined with getter-only @property decorators. For example, 
 ```
   >>> g = Gaussian(IIDStdUniform(2, seed=7), covariance=np.eye(2))
   >>> g.covariance[0, 0] = 9   # illegal user operation 
   >>> print(f"{g.covariance[0, 0] = }") 
    g.covariance[0, 0] = np.float64(9.0)
  ```

ishn29:
> 4. For  covariance matrices of `Uniform(IIDStdUniform(d))` and `Kumaraswamy(IIDStdUniform(d))` that have zeros in off-diagonal entries, we just need to store the diagonal vector to save memory consumption from O(d^2) to O(d).

@sou-cheng-choi could you clarify whether covariance should be stored as a separate diagonal vector? For Uniform and Kumaraswamy, that vector is identical to variance. Would it be preferable for covariance to reference variance, or should we remove the redundant attribute and use variance as the compact covariance representation?

ishn29:
- `mean`, `variance`, `standard_deviation`, and `covariance` now read only properties.
- Stored diagonal covariance matrices sparsely for `Uniform` and `Kumaraswamy`
- Compute variance for `Kumaraswamy` as `mean^2 * expm1(q)` with `q = K(2) - 2K(1)` built by quadrature of the log moment curvature K’’
- Added `Uniform` validation requiring finite bounds with `upper_bound > lower_bound`
- Added variance and standard deviation attributes to `BrownianMotion`.
- Fixed `MaternGP` spawning so kernel variance, nugget, mean, points, and other kernel parameters are preserved. Incompatible dimension changes now raise `DimensionError`.
- Demo notebook now loops over `Uniform` `Kumaraswamy` and `Gaussian`
- Added unit tests which numerically verify the mean and covariance matrix for the objects

alegresor (approved):
Looks great @ishn29! One final comment: Please add the reference where you found the  Kumaraswamy mean and variance if not available on the already-linked Wikipedia page

fjhickernell (changes requested):
ZeroInflatedExpUniform needs a mean, variance, etc.

fjhickernell:
I also notice that the mean and perhaps variance, etc. of univariate distributions are returned as 1-d arrays.  May they be returned as scalars? @alegresor what do you think?

would np.squeeze work?

fjhickernell (approved):
Looks good now

fjhickernell:
@sou-cheng-choi what is left?

ishn29:
Hi @sou-cheng-choi could you take a look and give me feedback. Thanks

sou-cheng-choi (changes requested):
@ishn29 

There are a few more edge cases. Could you help investigate?

1.  In `test_variance_matches_closed_form()` in `test/test_true_measures.py`, If I add one more test case as follows, the unit test fails. Please debug.

```{python}
        a = np.array([0.01, 1.0, 2.0, 3.5])
        b = np.array([1, 3.0, 4.0, 1.5])
        kumaraswamy = Kumaraswamy(DigitalNetB2(4, seed=7), a=a, b=b)
```

2. The following prints `mean: reported=-18.0, expected=110.0`. Please investigate.

```{python}
    lower = np.array([100], dtype=np.int8)
    upper = np.array([120], dtype=np.int8)
    uniform = Uniform(IIDStdUniform(1, seed=7), lower_bound=lower, upper_bound=upper)
    expected_mean = 110.0
    print(f"mean: reported={uniform.mean}, expected={expected_mean}")
```

3.   MaternGP.variance is changed from a scalar supplied by the user into a vector of marginal variances without a deprecation or renamed public replacement.  In QMCPy version 2.3, it was a scalar value. Please handle this issue. 

```{python}
       m = MaternGP(IIDStdUniform(3, seed=7),
        points=np.linspace(0, 1, 3)[:, None],
        variance=0.01)
      print(f"constructor variance=0.01; public variance={m.variance}")   
      # constructor variance=0.01; public variance=[0.010001 0.010001 0.010001]
``` 

4. `Kumaraswamy(IIDStdUniform(1), a=np.inf, b=np.inf)` returns mean and variance as `nan`. We probably want to make sure inputs `a` and `b` are finite and positive.

5. `Uniform(IIDStdUniform(10000)).covariance` still lgives a dense matrix with zero off-diagonal entries.

ishn29:
Hi @sou-cheng-choi, thank you for the feedback. I have fixed issues 1-4 and I could use some clarification on 5.

### 1. Kumaraswamy variance test failure (`a=0.01`)
rewrote the moments in closed form with the log Beta function. With `K(r) = log(b) + betaln(1 + r/a, b)`, we have `mean = exp(K(1))` and `variance = mean² · expm1(K(2) − 2·K(1))`. Since `K` is convex, `q = K(2) − 2K(1) >= 0`, so `expm1` computes `eq − 1` without cancellation. 

### 2. Uniform int8 mean returns −18 instead of 110
changed bounds to float64 at construction (`self.a = np.array(lower_bound, dtype=np.float64)`, same for `b`). Mean now is 110.

### 3. MaternGP.variance breaking change
The constructor stores the scalar as `_kernel_variance`, a new public `kernel_variance` property returns it, and accessing `.variance` gives a `DeprecationWarning` pointing to `kernel_variance`. So `m.variance` is marginal variance vector and `m.kernel_variance` is `0.01`.

### 4. Kumaraswamy accepting non-finite / non-positive `a`, `b`
the constructor now validates that `alpha` and `beta` are finite and strictly positive, raising `ParameterError("Kumaraswamy requires finite a,b>0.")` otherwise (also rejects `nan`, `0`, and negatives).

### 5. `Uniform(IIDStdUniform(10000)).covariance` still gives a dense matrix with zero off diagonal entries.

I had a quick question on what you'd like .covariance to return.

Right now the internal storage is O(d). I store the diagonal as a scipy.sparse dia_matrix. The output you showed above is the public .covariance property, which reconstructs a dense read only ndarray on access. I made it dense on purpose so the public API and printed output stay unchanged, since the existing tests expect .covariance to be a read only (d, d) ndarray (they call np.diag(...), assert_allclose(...), and check .flags.writeable on it).

So the memory savings still apply to what is stored persistently, but any single access to .covariance still creates the O(d²) dense form.

Do you want .covariance itself to return the sparse matrix directly or is O(d) storage with a dense on access property what you had in mind?

sou-cheng-choi:
@ishn29 

 If a matrix is sparse, we can simply display its spasre format without calling `toarray()` to change the matrix to dense format  as in qmcpy/true_measure/abstract_true_measure.py: line 76. Here is an example of a 3x3 diagonal matrix in sparse format:
 
 ```
  Coords  Values
  (0, 0)  0.08333333333333333
  (1, 1)  0.08333333333333333
  (2, 2)  0.08333333333333333
 ```

ishn29:
@sou-cheng-choi,

I just pushed an update returning diagonal covariance sparsely for Uniform and Kumaraswamy. Please let me know if there is anything else that needs to be changed.

Thanks

sou-cheng-choi (approved):

fjhickernell:
If this is ready to go, then please squash and merge

PR #580, opened 2026-08-01
Simplifying and Sorting QMCPy Imports. Removing Trailing Space.

sou-cheng-choi:
*   The core import structure is simplified by exposing key classes and utilities (e.g., `AcceptanceRejection`, `TriangularDistribution`) at the top-level `qmcpy` package.
*   Users can now directly import these components using `from qmcpy import ...`, eliminating deep submodule paths.
*   All demo notebooks, scripts, and documentation have been updated to utilize this new flat import style.
*   Developer experience is improved by adding a dedicated CI testing page (`docs/ci-testing.md`) summarizing GitHub Actions policies.
*   Makefile targets are enhanced with better comments and added functionality (e.g., `flatten_qmcpy_imports`).


fjhickernell (approved):
Quick question, what is our minimum Python level at the moment?  And is this the right level.

fjhickernell:
And thanks, Sou-Cheng.

qmcpy/stopping_criterion/__init__.py
```diff
@@ -24,10 +29,17 @@ def __init__(self, *args, **kwargs):
                 "PFGPCI requires torch and gpytorch but no installations found"
             )
 
+    class PFGPCIData(object):
+        def __init__(self, *args, **kwargs):
+            raise ModuleNotFoundError(
+                "PFGPCIData requires torch and gpytorch but no installations found"
+            )
```
alegresor, line 29: I don't think PFGPCIData needs to be available at the top level as we don't want users to instantiate it accidentally. It is a helper object used by PFGPCI

qmcpy/__init__.py
alegresor, line 1: I like the promotion of `AcceptanceRejection` and `AcceptanceRejectionReal` and `TriangularDistribution` to the top level. 

For the rest of the file, I do not think we should promote any of these utility methods to the top level. They are quite specific and I would think more confusing than useful to users. Users who need these methods can still access them with `from qmcpy.util import` what they need, but I would leave them where they are.

alegresor (changes requested):

qmcpy/__init__.py
sou-cheng-choi: Will revert.

sou-cheng-choi:
> Quick question, what is our minimum Python level at the moment? And is this the right level.

@fjhickernell 
`pyproject.toml` says `requires-python = ">= 3.5" `, but 

-  use of f-strings require at least Python 3.6
- torch >= 2.2 requires at least Python 3.8
- pytest >= 9.0.3 needs at least Python 3.10

I have open [issue 583](https://github.com/QMCSoftware/QMCSoftware/issues/583) to address this problem.

sou-cheng-choi:
@alegresor @fjhickernell 

Please re-review when you have a moment.

scripts/flatten_qmcpy_imports.py
```diff
@@ -0,0 +1,224 @@
...
+)
+NOTEBOOK_STAR_IMPORT_RE = re.compile(
+    rb'^(?P<json_indent>[ \t]*)"(?P<code_indent>(?:[ \t]|\\t)*)'
+    rb"from[ \t]+qmcpy[ \t]+import[ \t]+\*(?:\\r)?(?:\\n)?\""
+    rb",?[ \t]*(?:\r\n|\n|\r)?$"
+)
+
+
+def _star_import_context(line: bytes) -> tuple[bytes, bytes] | None:
+    text_match = TEXT_STAR_IMPORT_RE.match(line)
```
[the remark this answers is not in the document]
JiangruiKang: I agree this should block the PR until the supported-Python contract is made internally consistent. I also checked the Actions run directly: the job labeled `Unit Tests on macos-latest (Python 3.5)` actually reports `python 3.14.6-final-0` during pytest. The workflow defines `matrix.python-version`, but that value is not passed to `setup-miniconda`, so the current green matrix does not validate Python 3.5–3.14 compatibility.

I think we should make one explicit project-level decision here: either keep the declared old-Python support and make the tooling/tests plus CI genuinely compatible with it, or raise `requires-python` to the real supported minimum and align CI/dependencies/docs accordingly. I would not merge the new formatter until that contract is resolved.

qmcpy/__init__.py
```diff
@@ -35,7 +39,73 @@
...
+
+    _mpmc_utils_available = True
+    from .discrete_distribution.mpmc.models import MPMC_net
+except ImportError as error:
+    _missing_module = getattr(error, "name", None)
+    if _missing_module is None or _missing_module.split(".", 1)[0] not in {
+        "pyg_lib",
+        "torch",
+        "torch_geometric",
+    }:
```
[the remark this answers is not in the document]
JiangruiKang: I agree this remains blocking. Optional MPMC/PyG availability should not be able to make the base `import qmcpy` fail because one transitive optional dependency is absent.

At the same time, simply broadening the allowlist or catching every `ImportError` risks hiding genuine QMCPy import bugs. I would prefer a lazy optional export, or another guard that cleanly isolates optional MPMC dependencies from the base package import. Please add a regression test for a missing transitive PyG dependency as well, rather than only the direct `torch` / `torch_geometric` cases.

qmcpy/stopping_criterion/__init__.py
JiangruiKang: I agree. The current fallback actually makes this API inconsistent: `PFGPCIData` is defined when torch/gpytorch are missing, but it is not imported in the normal dependency-present branch. Since `PFGPCIData` is intended to remain an internal helper, I think the clean fix is to remove this fallback class rather than expose it conditionally through `stopping_criterion` / the top-level `qmcpy` namespace.

scripts/flatten_qmcpy_imports.py
```diff
@@ -0,0 +1,639 @@
...
+        return (
+            b"from"
+            + match.group("after_from")
+            + b"qmcpy"
+            + match.group("before_import")
+            + b"import"
+            + match.group("imported")
+        )
+
+    updated = QMCPY_IMPORT_RE.sub(replace, content)
```
JiangruiKang: **Blocking correctness issue:** this substitution operates on the raw file bytes, so it can rewrite text that is not an actual Python import statement.

For example:
```python
text = """
from qmcpy.integrand import Keister
"""
```
contains a substring that matches `QMCPY_IMPORT_RE`; that matched substring also parses successfully as an `ImportFrom`, so the formatter can change the string value to `from qmcpy import Keister`.

For `.py` / `.pyi` files, this transformation should operate on syntactic import nodes (or explicitly protect string/comment token spans) rather than applying the import regex to the entire raw byte stream. Please add a regression test showing that import-looking text inside strings/comments is preserved.

scripts/flatten_qmcpy_imports.py
```diff
@@ -0,0 +1,639 @@
...
+        elif isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):
+            bound.add(node.name)
+        elif isinstance(node, ast.arg):
+            bound.add(node.arg)
+        elif isinstance(node, ast.alias):
+            bound.add((node.asname or node.name).split(".")[0])
+        elif isinstance(node, ast.ExceptHandler) and node.name:
+            bound.add(node.name)
+
+    return (loaded & public_names) - bound
```
JiangruiKang, line 436: **Blocking correctness issue:** the file-wide `bound` set is not a safe approximation for expanding star imports because bindings from unrelated scopes — or assignments later in the file — can suppress imports that are still required at the point of use.

For example:
```python
from qmcpy import *

def f(Lattice):
    return Lattice

x = Lattice(dimension=2)
```
`Lattice` is added to `bound` because it is a parameter of `f`, so this analysis can omit `Lattice` from the rewritten import and leave the module-level use as a `NameError`.

There is a similar ordering problem with `x = Lattice(...)` followed by `Lattice = something_else`. I think star-import expansion needs scope- and statement-order-aware analysis, or it should conservatively leave star imports unchanged when that cannot be established safely. Please add regression tests for these cases.

scripts/flatten_qmcpy_imports.py
```diff
@@ -0,0 +1,639 @@
...
+def _load_qmcpy_public_names(repository_root: Path) -> frozenset[str] | None:
+    """Return qmcpy's public top-level names, or None if qmcpy can't be imported."""
+
+    if str(repository_root) not in sys.path:
+        sys.path.insert(0, str(repository_root))
+    try:
+        import qmcpy
+    except ImportError:
+        return None
+    return frozenset(name for name in dir(qmcpy) if not name.startswith("_"))
```
JiangruiKang: **Blocking determinism/API issue:** `dir(qmcpy)` is a runtime namespace, not a stable public-API specification. The top-level package imports modules whose exported names depend on optional dependencies, so two developers with different torch/gpytorch/MPMC environments can obtain different `public_names` and therefore different formatter output from the same source tree.

Formatting should be deterministic and based on an explicit/static API contract — e.g. maintained `__all__` declarations or a dedicated canonical export set — rather than importing the package and inspecting `dir(qmcpy)`.

scripts/remove_trailing_whitespace.py
```diff
@@ -0,0 +1,132 @@
...
+        and path.is_file()
+        and (path.name in SUPPORTED_NAMES or path.suffix.lower() in SUPPORTED_SUFFIXES)
+    )
+
+
+def _python_string_spans(text: str) -> dict[int, list[tuple[int, int | None]]]:
+    spans: dict[int, list[tuple[int, int | None]]] = {}
+    tokens = tokenize.generate_tokens(io.StringIO(text).readline)
+    for token in tokens:
+        if token.type != tokenize.STRING:
```
JiangruiKang: **Blocking semantics issue:** protecting only `tokenize.STRING` does not cover modern multiline f-strings. On current Python versions, f-strings are tokenized using `FSTRING_START` / `FSTRING_MIDDLE` / `FSTRING_END`, so trailing spaces inside the literal portion can be treated as ordinary source whitespace and removed, changing the runtime string value.

For example:
```python
x = f"""hello  
world  
"""
```
should preserve the two trailing spaces inside the string. Please extend the string-span handling to cover f-string literal spans and add a regression test; the current test only covers a regular triple-quoted string.

JiangruiKang (changes requested):
I reviewed the current head (`cc18288`) in more detail, including the formatter implementation and tests, optional-import behavior, and the CI configuration.

I found four additional correctness issues that I believe should block merging:

1. The nested-import rewrite operates on raw file bytes and can rewrite import-looking text inside Python string literals.
2. Star-import expansion uses file-wide name analysis that is not scope-aware or statement-order-aware and can remove required imports.
3. The formatter derives the public API from the runtime `dir(qmcpy)`, making formatting behavior dependent on installed optional dependencies rather than a stable API contract.
4. The trailing-whitespace formatter does not protect modern multiline f-string token spans and can therefore change runtime string values.

I have left inline comments with reproducible examples for these cases.

I also agree with the existing unresolved comments concerning the Python-version contract, optional MPMC imports, and `PFGPCIData`. In particular, I verified that the CI job labeled “Python 3.5” is actually running Python 3.14.6 because the matrix Python version is not passed to `setup-miniconda`.

The explicit-import cleanup itself looks useful, and I did not identify a numerical/QMC regression in the mostly whitespace/import-style changes. However, I do not think the new repo-wide formatting infrastructure is sufficiently semantics-preserving or deterministic yet.

Requesting changes.

sou-cheng-choi:
@JiangruiKang @alegresor 

I have made a few commits to address your comments. Please re-review when you have a moment.

JiangruiKang:
@fjhickernell The project currently declares Python >=3.5, but the existing CI does not actually verify that version. The job labeled “Python 3.5” ran Python 3.14 because the matrix version was not passed to setup-miniconda.

Based on the current package code, development tools, and test dependencies, I would recommend supporting Python 3.10 through 3.14 for now:

- declare `requires-python = ">=3.10"`
- run CI on Python 3.10, 3.11, 3.12, 3.13, and 3.14
- avoid setting an upper Python bound unless we know a newer version is incompatible

I would treat this as a separate project-level compatibility decision rather than changing it implicitly in this PR.

fjhickernell (approved):

fjhickernell:
> @fjhickernell The project currently declares Python >=3.5, but the existing CI does not actually verify that version. The job labeled “Python 3.5” ran Python 3.14 because the matrix version was not passed to setup-miniconda.
> 
> Based on the current package code, development tools, and test dependencies, I would recommend supporting Python 3.10 through 3.14 for now:
> 
> * declare `requires-python = ">=3.10"`
> * run CI on Python 3.10, 3.11, 3.12, 3.13, and 3.14
> * avoid setting an upper Python bound unless we know a newer version is incompatible
> 
> I would treat this as a separate project-level compatibility decision rather than changing it implicitly in this PR.

agree.  Shall we move the supported to 3.10 and above? @alegresor @sou-cheng-choi

JiangruiKang:
@sou-cheng-choi I have prepared local fixes for the remaining review issues:

- keep `PFGPCIData` internal rather than exposing it at the top level
- prevent the import formatter from modifying strings, comments, notebook metadata, or Markdown cells
- leave star-import expansion disabled because it is not scope/order safe
- restore the documented Makefile target and correct the whitespace-tool description
- add regression tests for these cases

The full unit suite passes locally: 537 passed, 1 skipped. I have not included any Python-version changes.

Would you prefer that I push these commits directly to `doc_tst_choi`, or push them to a separate branch for you to review/cherry-pick?

qmcpy/stopping_criterion/__init__.py
```diff
@@ -24,10 +28,17 @@ def __init__(self, *args, **kwargs):
...
```
alegresor: Please remove

qmcpy/__init__.py
```diff
@@ -22,7 +22,11 @@
     KernelMultiTask,
     KernelMultiTaskDerivs,
 )
-from .true_measure.acceptance_rejection import AcceptanceRejection, AcceptanceRejectionReal
+from .true_measure.acceptance_rejection import (
+    AcceptanceRejection,
+    AcceptanceRejectionReal,
+)
```
alegresor: These two should already be imported with the line `from .true_measure import *`.

qmcpy/__init__.py
```diff
@@ -22,7 +22,11 @@
...
+from .true_measure.triangular import TriangularDistribution
```
alegresor: This is a SciPy-compatible helper for the Triangular true measure. Don't think we want this at the top level.

alegresor (changes requested):

sou-cheng-choi:
> @sou-cheng-choi I have prepared local fixes for the remaining review issues:
> 
> * keep `PFGPCIData` internal rather than exposing it at the top level
> * prevent the import formatter from modifying strings, comments, notebook metadata, or Markdown cells
> * leave star-import expansion disabled because it is not scope/order safe
> * restore the documented Makefile target and correct the whitespace-tool description
> * add regression tests for these cases
> 
> The full unit suite passes locally: 537 passed, 1 skipped. I have not included any Python-version changes.
> 
> Would you prefer that I push these commits directly to `doc_tst_choi`, or push them to a separate branch for you to review/cherry-pick?

Sure. Please push it here when you have a moment.

sou-cheng-choi:
@alegresor  Thanks for catching those. I have removed them and fixed the associated import statements in demos and source files. Waiting for tests to pass. Please re-review.

@JiangruiKang Thanks for helping me fix the issues you have found. I have merged my latest changes with yours plus changes from `develop` branch. Waiting for tests to pass. Please re-review.

JiangruiKang (changes requested):
I re-reviewed the current head (`88c3feba`), including the fixes made after merging the latest `develop`.

The missing explicit imports introduced by the merge are now fixed, and `TriangularDistribution` is consistently kept out of the top-level API. The complete local unit suite passes: 634 passed, 1 skipped. I also rechecked the formatter safety tests, optional MPMC behavior, whitespace handling, and the documentation build; I did not identify a numerical or QMC algorithm regression.

I still found two issues that should be addressed before approval:

1. The repository is not clean under the formatter introduced by this PR. Running `python scripts/flatten_qmcpy_imports.py --check .` reports three files that would still be rewritten:
   - `qmcpy/discrete_distribution/digital_net_any_bases/hammersley.py`
   - `test/test_discrete_distribs.py`
   - `test/test_scipy_wrapper_custom.py`

   Since this PR introduces and applies the formatter repository-wide, the committed tree should be idempotent under its own check mode.

2. The merge removed the regression assertion that `PFGPCIData` is absent from the optional-free public-name set in `test_public_names_optional_free_stable`. The implementation currently keeps it internal, but that API decision is no longer protected by a test. Please restore `assert "PFGPCIData" not in names`. Since the latest change also intentionally keeps `TriangularDistribution` out of the top-level API, an analogous assertion for that name would make the contract explicit.

I am treating the Python-version declaration and CI matrix mismatch as a separate project-level issue, not as a blocker in this PR.

Requesting changes for the two items above.

sou-cheng-choi:
@JiangruiKang 

I have addessed Comments 1 and 2  in 

- lines 1-9 in `test_discrete_distribs.py` 
- line 7 in `est_scipy_wrapper_custom.py`, and 
- line3 in `hammersley.py`.
- lines 327-330 in `test_flatten_qmcpy_imports.py`.

Do you think these changes are sufficient?

sou-cheng-choi:
@alegresor  @JiangruiKang 

Need your re-re-review again.

JiangruiKang (approved):
Re-reviewed the current head (2e793f82). The two remaining issues from my previous review have been addressed: the committed tree is clean under the import formatter's check mode, and the public-API regression test now protects both PFGPCIData and TriangularDistribution from accidental top-level exposure. I also reviewed the subsequent Makefile and whitespace-only changes and did not find a functional or numerical regression. The complete local unit suite passed (634 passed, 1 skipped), and all GitHub checks are now passing. The Python-version contract remains a separate project-level discussion and is not a blocker for this PR. Approving.

alegresor (approved):

PR #585, opened 2026-08-08
Fix GBM demo replication statistics and speed up the Gaussian transform

larissensium:
Fixed a table in gbm_demo to correctly show results averaged over replications.


sou-cheng-choi (changes requested):
See my comments above

alegresor (approved):
Defer to @sou-cheng-choi  review on this

alegresor (approved):

sou-cheng-choi (changes requested):
@larissensium 

I have 

* merged in latest changes from develop branch
* corrected some typos
* made the paths from QuantLib independent with different seed values in `quantlib_util.py`
* added unit tests 

Please review and verify the changes.

demos/GBM/gbm_code/quantlib_util.py
```diff
@@ -48,18 +72,26 @@ def generate_quantlib_paths(
...
     elif sampler_type == "Sobol":
-        uniform_rsg = ql.UniformLowDiscrepancySequenceGenerator(dimension, seed)
-        sequence_gen = ql.GaussianLowDiscrepancySequenceGenerator(uniform_rsg)
-        path_gen = ql.GaussianSobolMultiPathGenerator(
-            gbm, list(times), sequence_gen, False
+        # UniformLowDiscrepancySequenceGenerator is deterministic for the fixed
+        # Jaeckel direction integers, so changing its `seed` does not create an
+        # independent replication. Burley2020SobolRsg applies a seeded Owen-style
+        # scramble; keep the underlying Sobol seed fixed and vary the scramble.
+        uniform_rsg = ql.Burley2020SobolRsg(dimension, 0, ql.SobolRsg.Jaeckel, seed)
```
JiangruiKang, line 79: [P2] Use a fixed nonzero underlying Sobol seed

QuantLib treats this second argument differently from the replication-specific scramble seed. Jaeckel direction integers are tabulated only through dimension 32; above that, `SobolRsg` uses this value to generate additional direction integers, and QuantLib defines `0` as a clock-selected random seed. Consequently, two calls with the same public `seed` can produce different paths when `n_steps >= 33` (see lballabio/QuantLib#2732), even though the comment says the underlying seed is fixed. Please use a fixed nonzero value such as `42`, vary only `scrambleSeed`, and add a regression test at `n_steps=33` showing that equal scramble seeds reproduce exactly while different scramble seeds remain distinct.

JiangruiKang (changes requested):
The switch to Burley scrambling is the right way to obtain independent randomized-QMC replications, but one reproducibility defect remains at dimensions above the Jaeckel table. The current tests are green because they use `n_steps=4`; please cover the 32/33-dimensional boundary before merging.

qmcpy/true_measure/gaussian.py
```diff
@@ -158,7 +159,9 @@ def mvn_scipy(self, value):
         self._mvn_scipy_cache = value
 
     def _transform(self, x):
-        return self.mu + np.einsum("...ij,kj->...ik", norm.ppf(x), self.a)
+        transformed = ndtri(x) @ self.a.T
```
alegresor, line 162: Nice change! Haven't heard of this `ndtri` before, but it is suppose to be much faster than `norm.ppf`

alegresor (approved):

JiangruiKang (approved):
The QuantLib Sobol reproducibility issue above 32 dimensions has been addressed with a fixed nonzero direction seed and regression coverage. The current head is green, and the targeted GBM checks pass locally.

sou-cheng-choi (approved):

PR #593, opened 2026-08-12
Update QMCPy/__init__.py with version 2.4 and remove top-level __init__.py

sou-cheng-choi:


alegresor (approved):

rvare (approved):
Everything appears to be in order.

PR #608, opened 2026-08-19
Add summary statistics to ProductMeasure

Laasya-73:
This PR adds combined summary statistics to `ProductMeasure`.

`ProductMeasure` now derives statistics for its independent marginal blocks by:

- concatenating marginal means
- concatenating marginal variances
- concatenating marginal standard deviations
- constructing a block-diagonal covariance matrix from the marginal covariance matrices

This allows users to access statistics for the complete product distribution directly instead of calculating them separately for each marginal.

### Changes

- Added combined summary-statistic support to `ProductMeasure`.
- Added block-diagonal covariance construction for independent marginal blocks.
- Added tests for one-dimensional and multidimensional marginal combinations.
- Added coverage for scalar one-dimensional statistics.
- Added tests for behavior when marginal statistics are unavailable.

### DIFF
--- a/qmcpy/true_measure/product_measure.py
+++ b/qmcpy/true_measure/product_measure.py
@@ -5,7 +5,7 @@ from .abstract_true_measure import AbstractTrueMeasure
 from ..discrete_distribution.abstract_discrete_distribution import (
     AbstractDiscreteDistribution,
 )
-from ..util import DimensionError, ParameterError
+from ..util import DimensionError, ParameterError, _univ_repr
 
 
 class ProductMeasure(AbstractTrueMeasure):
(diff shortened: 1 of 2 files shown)


alegresor (approved):

fjhickernell (approved):
tried it with the AreWeThereYet notebook for MATH 565 and it worked like a charm.

sou-cheng-choi (changes requested):
See example above.

sou-cheng-choi (changes requested):
@Laasya-73 

Just a few loose ends to tie up here: 

1. Consider using 'csr' instead of 'dia' in line 253 of `qmcpy/true_measure/product_measure.py` for more efficient storage of general, irregular sparsity, and use `dia' for a diagonal/banded matrix.

2. Do we need to print the whole covariance matrix in `repr` or shall we just print the header or a compact summary like "sparse CSR, shape=(130,130), nnz=132"?

3. Suggest to implement a lazy cache mechanism to improve `ProductMeasure`, which currently rebuilds marginal statistics (e.g., mean, variance) from scratch upon every access. Add a quick unit tests to confirm the cache enables substantially faster second access. An example of lazy cache in QMCPy is found in `gaussian.py`.

sou-cheng-choi (approved):

PR #620, opened 2026-09-13
Speedup QMCPy

sou-cheng-choi:
See demo [`demos/performance_optimizations_demo.ipynb`](https://github.com/QMCSoftware/qmcpy/blob/speedup_choi/demos/performance_optimizations_demo.ipynb)


qmcpy/integrand/financial_option.py
```diff
@@ -574,7 +574,7 @@ def get_exact_value(self):
                     np.log(self.start_price / self.strike_price)
                     + (self.interest_rate - self.volatility**2 / 2) * self.t_final
                 )
-                fp = self.start_price * norm.cdf(term1 / denom) - decay * norm.cdf(
+                fp = self.start_price * ndtr(term1 / denom) - decay * ndtr(
```
alegresor, line 577: I like the `ndtr` and `ndtri` updates

qmcpy/integrand/bayesian_lr_coeffs.py
```diff
@@ -78,7 +78,7 @@ def __init__(
         )
 
     def g(self, x):
-        z = np.einsum("...j,ij->...i", x, self.feature_array)
+        z = x @ self.feature_array.T
```
alegresor, line 81: `x` may be three dimensional, in which case I think this will break. Also, I think `np.einsum` is usually as fast as possible, so maybe this line doesn't need to be changed?

qmcpy/true_measure/brownian_motion.py
alegresor, line 1: What has been changed here? Something new added?

qmcpy/true_measure/gaussian_copula.py
```diff
@@ -120,8 +120,8 @@ def _weight(self, x):
         except ParameterError:
             return self._unit_weight_with_warning(x)
 
-        z = norm.ppf(u)
-        quad = np.einsum("...i,ij,...j->...", z, self._corr_inv_minus_eye, z)
+        z = ndtri(u)
+        quad = ((z @ self._corr_inv_minus_eye) * z).sum(-1)
```
alegresor, line 124: Same comment as earlier, I think this will break if `z` is three dimensional.

qmcpy/true_measure/geometric_brownian_motion.py
```diff
@@ -149,31 +141,9 @@ def _compute_gbm_covariance(self):
...
-                    cov_ij = (
-                        S0_sq
-                        * exp_mu_t[i]
-                        * exp_mu_t[j]
-                        * (exp(self.diffusion * t_min_ij) - 1)
-                    )
-                    cov_matrix[i, j] = cov_ij
-                    cov_matrix[j, i] = cov_ij  # Symmetric
-
-        return cov_matrix
```
alegresor, line 176: No need to split this into small + large cases anymore? More efficient to just use simplified construction?

alegresor (changes requested):

qmcpy/integrand/bayesian_lr_coeffs.py
sou-cheng-choi: It does not seem to break in my tests cases with 3D or higher dimensions. Could you shoe me cases that it will break if you are aware of any? 

Example 5 in the demo show that the change could be 1.7 times faster.

qmcpy/true_measure/gaussian_copula.py
sou-cheng-choi: Same as my previous resposne.

qmcpy/true_measure/brownian_motion.py
sou-cheng-choi: 1. The 1_transform1 function got a one-line swap: `norm.ppf(x)` -> `ndtri(x)`.
2. The new `_bridge_transform` improves performance by precomputing node groupings by depth in Owen's tree.  This optimization reduces the computational complexity from $O(d)$ interpreter steps to $O(log d)$ for nodes at equal depth. See Example 2 in demo notebook please.

qmcpy/true_measure/geometric_brownian_motion.py
sou-cheng-choi: Right, it seems so. It is cleaner and faster, though there is some memory trade off, which I think is not a major issue until proven otherwise.

qmcpy/integrand/bayesian_lr_coeffs.py
alegresor: You're right, this would only break if x is 1d, which should never be the case.

alegresor (approved):
Thanks for carefully addressing my concerns!

qmcpy/discrete_distribution/digital_net_any_bases/digital_net_any_bases.py
```diff
@@ -208,7 +212,7 @@ def __init__(self,
...
+        self.all_primes = _ALL_PRIMES
```
fjhickernell, line 215: `_ALL_PRIMES` is a writable module-level NumPy array, so assigning it directly makes every Halton instance share the same mutable object. I reproduced that mutating one instance’s `all_primes` changes the bases of subsequently created instances; previously each instance owned its array. Please use `_ALL_PRIMES.copy()` (or otherwise preserve per-instance behavior) and add a regression test demonstrating that instances are independent.

fjhickernell (changes requested):
Thanks for this substantial performance work. I am requesting changes for the shared mutable Halton prime table noted inline.

Before approval, please also:

- Revise the performance notebook’s Brownian-bridge claim: batching reduces Python-level loop iterations to O(log d), but the total numerical work remains O(d).
- Complete the PR description with the associated issue, scope/API impact, exact validation commands, and the AI-assistance and independent-verification disclosure required by the current contribution policy.

Validation performed: GitHub CI is currently green; `git diff --check` passed; and the focused local suite passed with 99 tests and 34 subtests.

sou-cheng-choi:
@fjhickernell 

I really appreciate your careful review. I have updated the PR description and made a few commits to

- address the mutable issue related to _ALL_PRIMES and wrong complexity you identified
- speed up notebook tests and improved doctests in CI.

Kindly re-review when you have a moment.

